PR #2, opened 2020-02-20
Update sql.js

lovasoa:
See https://github.com/cdnjs/cdnjs/issues/13685

### DIFF
--- a/packages/s/sql.js.json
+++ b/packages/s/sql.js.json
@@ -1,7 +1,6 @@
 {
   "name": "sql.js",
   "filename": "js/sql.js",
-  "version": "1.1.0",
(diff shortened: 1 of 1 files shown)


packages/s/sql.js.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sql.js",
   "filename": "js/sql.js",
-  "version": "0.5.0",
+  "version": "1.1.0",
```
xtuc: I think the version can be removed from here
```suggestion
```

xtuc:
I noticed (https://github.com/cdnjs/packages/issues/4) that the dist folder is preserved, we should instead specify dist in basePath

PR #36, opened 2020-05-06
Add jquery.lifestream w/ npm auto-update

MattIPv4:
Adding jquery.lifestream using npm auto-update from NPM package jquery-lifestream.

Resolves https://github.com/cdnjs/cdnjs/issues/12449.

### DIFF
--- a/packages/j/jquery.lifestream.json
+++ b/packages/j/jquery.lifestream.json
@@ -1,0 +1,30 @@
+{
+  "name": "jquery.lifestream",
+  "description": "jQuery plugin to show a stream of your online activity.",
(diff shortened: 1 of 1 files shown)


robocdnjs:
**Tests failed:**
```text
packages/j/jquery.lifestream.json: error: package download per month on npm is under 800

```

xtuc:
I think that given that the name is pretty long, it won't collide with a package that's more popular. do you think we should merge it still?

MattIPv4:
We already have it, it just doesn't have an auto-update config.

https://cdnjs.com/libraries/jquery.lifestream

PR #39, opened 2020-05-06
Add bchaddrjs w/ npm auto-update

MattIPv4:
Adding bchaddrjs using npm auto-update from NPM package bchaddrjs.

Resolves https://github.com/cdnjs/cdnjs/issues/12475.

### DIFF
--- a/packages/b/bchaddrjs.json
+++ b/packages/b/bchaddrjs.json
@@ -1,0 +1,35 @@
+{
+  "name": "bchaddrjs",
+  "description": "Bitcoin Cash general purpose address translation.",
+  "keywords": [
+    "bitcoin-cash",
+    "bitcoin",
+    "bch",
+    "cryptocurrency",
+    "address",
+    "encoding",
+    "translation",
+    "base58",
+    "bitpay",
+    "cash-address"
+  ],
+  "author": {
+    "name": "Emilio Almansi",
+    "email": "hi@ealmansi.com"
+  },
+  "license": "MIT",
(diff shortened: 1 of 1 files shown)


robocdnjs:
**Tests passed.**
Preview for `packages/b/bchaddrjs.json`
Last version (0.4.8):
```
bchaddrjs-0.4.8.js
bchaddrjs-0.4.8.min.js
```
10 Last version:
```
0.4.8: 2 file(s) matched
0.4.7: 2 file(s) matched
0.4.6: 2 file(s) matched
0.4.5: 2 file(s) matched
0.4.4: 2 file(s) matched
0.4.3: 2 file(s) matched
0.4.2: 2 file(s) matched
0.4.1: 2 file(s) matched
0.4.0: 2 file(s) matched
0.3.2: 2 file(s) matched
```

packages/b/bchaddrjs.json
```diff
@@ -0,0 +1,35 @@
...
+  "npmName": "bchaddrjs",
+  "npmFileMap": [
+    {
+      "basePath": "dist",
+      "files": [
+        "*.js"
+      ]
+    }
+  ]
+}
```
xtuc, line 35: there's no filename. I believe it will only have an impact on the website where we don't show the default?

should we warn?

MattIPv4: Yeah, filename is only used to highlight a default file. This package has the file names with versions in them, so there isn't anything I can set as the default.

I think warning would be okay, but it shouldn't block the PR passing.

PR #178, opened 2020-05-25
Add jsxgraph w/ npm auto-update

alfredwassermann:
JSON file for auto-update of JSXGraph via npm.

### DIFF
--- a/packages/j/jsxgraph.json
+++ b/packages/j/jsxgraph.json
@@ -33,8 +33,5 @@
       ]
     }
   ],
-  "licenses": [
-    "MIT",
-    "LGPL-3.0"
-  ]
+  "license": "MIT"
 }
--- a/packages/j/jsxgraph.json
+++ b/packages/j/jsxgraph.json
@@ -24,7 +24,12 @@
     {
       "basePath": "distrib",
       "files": [
-        "**/*.+(js|css)"
+        "jsxgraphcore.js",
+        "jsxgraph.css",
+        "geonext.min.js",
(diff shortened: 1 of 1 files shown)


MattIPv4:
@xtuc could you just double-check this before I merge as the CI can't?

packages/j/jsxgraph.json
```diff
@@ -0,0 +1,32 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/jsxgraph/jsxgraph.git"
+  },
+  "npmName": "jsxgraph",
+  "npmFileMap": [
+    {
+      "basePath": "distrib",
+      "files": [
+        "**/*.+(js|css)"
```
MattIPv4, line 27: This is going to include a lot of assets from the docs, which we don't want?

alfredwassermann: Sorry, I will postpone that after cleaning the npm repository with the next release.

xtuc:
The CI won't pass because the packge is not popular enough but I'm fine with merging

packages/j/jsxgraph.json
MattIPv4: We will still pull old versions though, using this fileMap, so this will still match assorted docs files from previous versions. Can we update the fileMap so it only matches the required files please?

MattIPv4: I recommend using https://www.digitalocean.com/community/tools/glob and importing the NPM package, so you can see what the glob would match.

PR #267, opened 2020-06-19
Update startbootstrap-sb-admin-2 w/ git auto-update

VenatusSimpleX:
The site's version is apparently stuck at 3.3.7+1 as shown in the screen shot attached. After checking I figure that it was due to the specific version being pinned, so I remove `version` in hopes that the cdn will update the template.

First time here.. sorry for any mistakes made.

Screenshot:
image

### DIFF
--- a/packages/s/startbootstrap-sb-admin-2.json
+++ b/packages/s/startbootstrap-sb-admin-2.json
@@ -13,7 +13,7 @@
     "target": "git://github.com/BlackrockDigital/startbootstrap-sb-admin-2.git",
     "fileMap": [
       {
-        "basePath": "dist",
+        "basePath": ".",
         "files": [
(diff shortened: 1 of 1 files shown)


packages/s/startbootstrap-sb-admin-2.json
```diff
@@ -14,7 +13,7 @@
     "target": "git://github.com/BlackrockDigital/startbootstrap-sb-admin-2.git",
     "fileMap": [
       {
-        "basePath": "dist",
+        "basePath": ".",
```
MattIPv4, line 16:
```suggestion
        "basePath": "",
```
Empty string is the same as `.`, but consistent with existing packags.

VenatusSimpleX:
Ah, thanks for the edit. Sorry for that mistake.

Issue #270, opened 2020-06-20
Add Library“rax” 


imsobear:
**Library name:** rax

**Library license (include link):** [BSD](https://github.com/alibaba/rax/blob/master/LICENSE)

**Library source (NPM package or Git repo):** https://www.npmjs.com/package/rax

**Official Homepage:** https://rax.js.org/

PR #271, opened 2020-06-20
Add rax w/ npm auto-update

imsobear:
Adding rax using npm auto-update from rax.

Resolves https://github.com/cdnjs/packages/issues/270.

### DIFF
--- a/packages/r/rax.json
+++ b/packages/r/rax.json
@@ -12,15 +12,18 @@
     "jsx"
   ],
   "homepage": "https://rax.js.org/",
-  "npmName": "rax",
-  "npmFileMap": [
-    {
-      "basePath": "dist",
-      "files": [
-        "**/*"
-      ]
-    }
-  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "rax",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
+        ]
+      }
+    ]
+  },
   "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


packages/r/rax.json
```diff
@@ -0,0 +1,29 @@
...
+  "homepage": "https://rax.js.org/",
+  "npmName": "rax",
+  "npmFileMap": [
+    {
+      "basePath": "dist",
+      "files": [
+        "**/*"
+      ]
+    }
+  ],
```
MattIPv4, line 23:
```suggestion
  "autoupdate": {
    "source": "npm",
    "target": "rax",
    "fileMap": [
      {
        "basePath": "dist",
        "files": [
          "**/*"
        ]
      }
    }
  ],
```
Please use the new `autoupdate` format documented in CONTRIBUTING.md :)

imsobear: Fixed.

packages/r/rax.json
```diff
@@ -0,0 +1,32 @@
...
+  "homepage": "https://rax.js.org/",
+  "autoupdate": {
+    "source": "npm",
+    "target": "rax",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 22:
```suggestion
          "*.js?(.map)"
```
cdnjs doesn't support hosting `.mjs` files, so we should only include the cjs & umd bundled files (and their maps).

PR #279, opened 2020-06-24
Update rambda w/ npm auto-update

benhormann:
Rambda is outdated, the npm package no longer has a minified `webVersion.js` bundle.

It has slightly broken the web app, since it thinks the latest version is one that hasn't got a corresponding folder.

### DIFF
--- a/packages/r/rambda.json
+++ b/packages/r/rambda.json
@@ -1,8 +1,8 @@
 {
   "name": "rambda",
-  "filename": "webVersion.js",
-  "version": "1.2.2",
+  "filename": "rambda.umd.js",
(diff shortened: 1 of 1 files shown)


packages/r/rambda.json
```diff
@@ -18,13 +18,16 @@
...
-    }
-  ]
+  "autoupdate": {
+    "source": "npm",
+    "target": "rambda",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "*.js?(.map)"
```
MattIPv4, line 28:
```suggestion
          "*.js"
```
There are no map files provided in the package, so there is no point in adding that to the glob?

PR #280, opened 2020-06-25
Add vanta w/ npm auto-update

dinedal:


packages/v/vanta.json
```diff
@@ -0,0 +1,33 @@
...
+    "three",
+    "threejs",
+    "background",
+    "canvas",
+    "javascript",
+    "webgl"
+  ],
+  "author": "Teng Bao",
+  "license": "MIT",
+  "npmName": "vanta",
```
MattIPv4, line 20:
```suggestion
```
We don't need the `npmName` property, the `autoupdate` property replaces it

packages/v/vanta.json
```diff
@@ -0,0 +1,33 @@
...
+  "autoupdate": {
+    "source": "npm",
+    "target": "vanta",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 28:
```suggestion
          "*.js"
```
We only need the `js` files from `dist`, not anything else that might get added later on.

PR #281, opened 2020-06-25
Add rambdax w/ npm auto-update

benhormann:
This is the extended version of Rambda. It has thousands of weekly downloads <https://www.npmjs.com/package/rambdax>, though the stars are a quite a bit lower than Rambda.

I'm not a maintainer, if that in any way helps the case for adding this package.

### DIFF
--- a/packages/r/rambdax.json
+++ b/packages/r/rambdax.json
@@ -14,7 +14,7 @@
     "util"
   ],
   "author": {
-    "name": "self_refactor",
+    "name": "Dejan Toteff",
     "url": "https://github.com/selfrefactor"
   },
   "license": "MIT",
--- a/packages/r/rambdax.json
+++ b/packages/r/rambdax.json
@@ -1,0 +1,33 @@
+{
+  "name": "rambdax",
+  "filename": "rambdax.umd.js",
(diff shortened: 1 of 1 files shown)


packages/r/rambdax.json
```diff
@@ -0,0 +1,33 @@
...
+  },
+  "license": "MIT",
+  "autoupdate": {
+    "source": "npm",
+    "target": "rambdax",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "*.js?(.map)"
```
MattIPv4, line 28:
```suggestion
          "*.js"
```
There are no `map` files, we just need the `js` files :)

packages/r/rambdax.json
```diff
@@ -0,0 +1,33 @@
...
+    "url": "git+https://github.com/selfrefactor/rambdax.git"
+  },
+  "keywords": [
+    "rambda",
+    "lodash",
+    "ramda",
+    "util"
+  ],
+  "author": {
+    "name": "self_refactor",
```
MattIPv4, line 17:
```suggestion
    "name": "Dejan Toteff",
```
We can use the author's actual name

PR #285, opened 2020-06-26
Test CI Checker for Oversize Files

tc80:
Fake package to test CI checker to see if it warns against files > 10MiB.


xtuc:
do we still need this?

PR #305, opened 2020-07-04
Remove airbrake-js git auto-update config

klausenbusk:
The auto-update config is out-of-date and the "version counter" has been
reset[1]. We can't overwrite existing versions, so we have no other
options.

[1] https://github.com/airbrake/airbrake-js/issues/602#issuecomment-545057436

Refs #302

### DIFF
--- a/packages/a/airbrake-js.json
+++ b/packages/a/airbrake-js.json
@@ -15,17 +15,5 @@
   "author": "Airbrake",
   "license": "MIT",
   "homepage": "https://airbrake.io",
-  "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/airbrake/airbrake-js.git",
-    "fileMap": [
-      {
-        "basePath": "dist",
-        "files": [
-          "client.*"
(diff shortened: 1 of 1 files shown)


MattIPv4:
I'm unsure about this, am curious what you think @xtuc?

Personally, I'd probably just want to update the auto-update config to be valid and let the bot pick up new versions when they don't collide with old ones.

klausenbusk:
> Personally, I'd probably just want to update the auto-update config to be valid and let the bot pick up new versions when they don't collide with old ones.

That `airbrake-js` package is [deprecated](https://www.npmjs.com/package/airbrake-js). Wouldn't it be better to deprecate this package and add a new `airbrake` or `@airbrake/browser` package?

MattIPv4:
Ah right, I had assumed `@airbrake/browser` was just a continuation of `airbrake-js` under a new name with reset versioning, but I think it probably would be sensible to add `@airbrake/browser` as a new package instead 👍

PR #308, opened 2020-07-04
Update clmtrackr w/ git auto-update

klausenbusk:
The auto-update glob is out-of-date, so let's switch to NPM and fix the
glob.

Refs #302

### DIFF
--- a/packages/c/clmtrackr.json
+++ b/packages/c/clmtrackr.json
@@ -14,11 +14,11 @@
     "url": "https://github.com/auduno/clmtrackr.git"
   },
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/auduno/clmtrackr.git",
+    "source": "npm",
+    "target": "clmtrackr",
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "build",
         "files": [
           "clmtrackr*"
         ]
--- a/packages/c/clmtrackr.json
+++ b/packages/c/clmtrackr.json
@@ -14,8 +14,8 @@
     "url": "https://github.com/auduno/clmtrackr.git"
   },
   "autoupdate": {
-    "source": "npm",
-    "target": "clmtrackr",
+    "source": "git",
+    "target": "git://github.com/auduno/clmtrackr.git",
     "fileMap": [
       {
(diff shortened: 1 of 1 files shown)


klausenbusk:
Hmm, the CI shouldn't fail if the file already exists.

MattIPv4:
Agreed

packages/c/clmtrackr.json
```diff
@@ -14,11 +14,11 @@
     "url": "https://github.com/auduno/clmtrackr.git"
   },
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/auduno/clmtrackr.git",
+    "source": "npm",
+    "target": "clmtrackr",
```
MattIPv4, line 18:
```suggestion
    "source": "git",
    "target": "git://github.com/auduno/clmtrackr.git",
```
The [git repo has the versions tagged correctly](https://github.com/auduno/clmtrackr/tree/v1.1.2/build), let's not complicate the change by updating where we get the files from.

klausenbusk:
The switch to NPM was done on purpose, as checking for newer versions 
should be faster (a few (one?) http call vs git clone). Don't we want that? :)

MattIPv4:
We discussed this internally -- unless we need to switch to make the package work, we'd much prefer to stay with the original source that the author chose for auto-updating. We don't have insight into why they chose Git over NPM, but there may be a reason. It also reduces the changes in the diff & reduces the risk of switching to an incorrect package, which has happened in a couple of the other PRs.

klausenbusk:
Fair enough :) Should we update the doc and state that NPM is preferred?

BTW: Do you use a tool to generate the glob url?

MattIPv4:
> Should we update the doc and state that NPM is preferred?

I don't think so, we're happy with either being used. Git cloning is a bit slower for sure, but that's not our biggest worry really. I'm happy for folks to add packages with either as a source.

> Do you use a tool to generate the glob url?

I built https://www.digitalocean.com/community/tools/glob a while back for testing globs easily on cdnjs, and even added a feature to quickly pull in files from an NPM package to make the whole process that much faster.

The CI generates glob URLs that go to that tool as well.

PR #310, opened 2020-07-04
Bulk 1: Update X w/ npm auto-update

klausenbusk:
This was done by-hand with help from `curl -s $(npm view X dist.tarball) | tar -tvzf -`

Refs #302

### DIFF
--- a/packages/s/slim.js.json
+++ b/packages/s/slim.js.json
@@ -15,13 +15,13 @@
     "binding"
   ],
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/eavichay/slim.js.git",
+    "source": "npm",
+    "target": "slim.js",
     "fileMap": [
       {
         "basePath": "",
         "files": [
-          "Slim*"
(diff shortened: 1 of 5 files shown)


MattIPv4:
I would much prefer these changes as a single JSON file per PR, please, so that we can effectively review each one and maintain a clear repo history/blame.

klausenbusk:
>  I would much prefer these changes as a single JSON file per PR, please, so that we can effectively review each one and maintain a clear repo history/blame.

Why would the history/blame be less clear? Are you squashing the commits?

MattIPv4:
Yes, all commits to master are squashed, so the general practice is one PR / one commit to master for each package change :)

klausenbusk:
Do you want me to split this PR?

MattIPv4:
Yes, if you could, please -- so that we have a PR for each package being modified. Thank you :)

klausenbusk:
Superseded by: #311, #312, #313, #314 and #315

PR #313, opened 2020-07-04
Update spin.js w/ npm auto-update

klausenbusk:
The auto-update glob is out-of-date, so let's switch to NPM and fix the
glob.

Refs #302

---
This was done by-hand with help from `curl -s $(npm view X dist.tarball) | tar -tvzf -`

### DIFF
--- a/packages/s/spin.js.json
+++ b/packages/s/spin.js.json
@@ -19,8 +19,7 @@
       {
         "basePath": "",
         "files": [
-          "spin.js",
-          "spin.css"
+          "*.@(js|css|ts)"
         ]
       }
     ]
--- a/packages/s/spin.js.json
+++ b/packages/s/spin.js.json
@@ -13,14 +13,14 @@
     "url": "git://github.com/fgnass/spin.js.git"
   },
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/fgnass/spin.js.git",
+    "source": "npm",
+    "target": "spin.js",
     "fileMap": [
       {
         "basePath": "",
         "files": [
           "spin.js",
-          "spin.min.js"
+          "spin.css"
         ]
       }
     ]


packages/s/spin.js.json
```diff
@@ -13,14 +13,14 @@
...
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/fgnass/spin.js.git",
+    "source": "npm",
+    "target": "spin.js",
     "fileMap": [
       {
         "basePath": "",
         "files": [
           "spin.js",
```
MattIPv4, line 22:
```suggestion
          "*.@(js|css|ts)",
```
There are also some typings included in the package, and as we support those, we might as well bring them in. [Glob](https://www.digitalocean.com/community/tools/glob?comments=true&glob=%2A.%40%28js%7Ccss%7Cts%29&matches=false&tests=%2F%2F%20Imported%20from%20spin.js%20NPM%20package&tests=spin.css&tests=spin.js&tests=package.json&tests=CHANGELOG.md&tests=LICENSE.md&tests=README.md&tests=spin.d.ts&tests=SpinnerOptions.d.ts)

packages/s/spin.js.json
```diff
@@ -13,14 +13,14 @@
...
-          "spin.min.js"
+          "spin.css"
```
MattIPv4, line 23:
```suggestion
```
Will be made redundant by the above glob

packages/s/spin.js.json
klausenbusk: Does people include `*.ts` directly on a webpage?

MattIPv4: Typings aren't used on web pages, but folks that are using Libman or Repl.it for example may want to be able to access typings. We support it, so we might as well include them if they're provided.

PR #316, opened 2020-07-04
Update regression w/ git auto-update

klausenbusk:

Refs #302

---

### DIFF
--- a/packages/r/regression.json
+++ b/packages/r/regression.json
@@ -16,13 +16,13 @@
     "analysis"
   ],
   "autoupdate": {
-    "source": "npm",
-    "target": "regression",
+    "source": "git",
+    "target": "git://github.com/Tom-Alexander/regression-js.git",
     "fileMap": [
       {
         "basePath": "dist",
         "files": [
-          "**/*"
+          "regression*.js"
         ]
       }
     ]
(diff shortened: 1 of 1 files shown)


packages/r/regression.json
```diff
@@ -16,13 +16,13 @@
     "analysis"
   ],
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/Tom-Alexander/regression-js.git",
+    "source": "npm",
+    "target": "regression",
```
MattIPv4, line 20:
```suggestion
    "source": "git",
    "target": "git://github.com/Tom-Alexander/regression-js.git",
```
The [git repo still has the correct files with tags](https://github.com/Tom-Alexander/regression-js/tree/2.0.1/dist), so let's not complicate the change by changing where we get files from.

packages/r/regression.json
```diff
@@ -16,13 +16,13 @@
...
     "fileMap": [
       {
-        "basePath": "build",
+        "basePath": "dist",
         "files": [
-          "regression*.js"
+          "**/*"
```
MattIPv4, line 25:
```suggestion
          "regression*.js"
```
The existing glob matches the files needed, `**/*` is far more generic and leaves us open to potential abuse.

PR #317, opened 2020-07-04
Update circliful w/ git auto-update

klausenbusk:

Refs #302

---

### DIFF
--- a/packages/c/circliful.json
+++ b/packages/c/circliful.json
@@ -1,19 +1,19 @@
 {
   "name": "circliful",
-  "filename": "js/jquery.circliful.min.js",
+  "filename": "circliful.js",
   "repository": {
     "type": "git",
-    "url": "https://github.com/pguso/jquery-plugin-circliful.git"
+    "url": "https://github.com/pguso/js-plugin-circliful.git"
   },
   "license": "MIT",
   "autoupdate": {
-    "source": "npm",
-    "target": "js-plugin-circliful",
+    "source": "git",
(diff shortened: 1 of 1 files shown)


packages/c/circliful.json
```diff
@@ -7,14 +7,13 @@
   },
   "license": "MIT",
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/pguso/jquery-plugin-circliful.git",
+    "source": "npm",
+    "target": "js-plugin-circliful",
```
MattIPv4, line 11:
```suggestion
    "source": "git",
    "target": "git://github.com/pguso/jquery-plugin-circliful.git",
```
The [git repo actually has more recent versions](https://github.com/pguso/js-plugin-circliful/tree/2.0.17/dist) than the [NPM package](https://www.npmjs.com/package/js-plugin-circliful), so let's stick to the git repo for the source-of-truth.

packages/c/circliful.json
```diff
@@ -7,14 +7,13 @@
...
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "dist",
         "files": [
-          "js/jquery.circliful*.js",
-          "css/jquery.circliful*.css"
+          "**/*"
```
MattIPv4, line 16:
```suggestion
          "*.@(js|css)"
```
`**/*` leaves us rather open to abuse, so we can use a more specific glob just for the files we need.

klausenbusk:
> Few more things on this one I've just realised, sorry:

Fixed..

PR #319, opened 2020-07-04
Update atmosphere w/ npm auto-update

klausenbusk:

Refs #302

---

### DIFF
--- a/packages/a/atmosphere.json
+++ b/packages/a/atmosphere.json
@@ -12,11 +12,11 @@
     "longpolling"
   ],
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/Atmosphere/atmosphere-javascript.git",
+    "source": "npm",
+    "target": "atmosphere.js",
     "fileMap": [
       {
-        "basePath": "modules/javascript/src/main/webapp/javascript",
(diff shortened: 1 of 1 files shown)


packages/a/atmosphere.json
```diff
@@ -12,11 +12,11 @@
...
-    "source": "git",
-    "target": "git://github.com/Atmosphere/atmosphere-javascript.git",
+    "source": "npm",
+    "target": "atmosphere.js",
     "fileMap": [
       {
-        "basePath": "modules/javascript/src/main/webapp/javascript",
+        "basePath": "lib",
         "files": [
           "atmosphere.js"
```
MattIPv4, line 21:
```suggestion
          "*.js"
```
There's also another `js` file in the package, and the docs don't really cover which should be used, so it's probably best for us to include both to be safe. [Glob](https://www.digitalocean.com/community/tools/glob?comments=true&glob=lib%2F%2A.js&matches=false&tests=%2F%2F%20Imported%20from%20atmosphere.js%20NPM%20package&tests=LICENSE&tests=lib%2Fatmosphere.js&tests=lib%2Fbrowser.js&tests=index.js&tests=package.json&tests=README.md)

PR #320, opened 2020-07-04
Update buckets w/ npm auto-update

klausenbusk:

Refs #302

---

### DIFF
--- a/packages/b/buckets.json
+++ b/packages/b/buckets.json
@@ -28,9 +28,9 @@
     "target": "git://github.com/mauriciosantos/Buckets-JS.git",
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "dist",
         "files": [
-          "buckets*.js"
+          "buckets?(.min).js?(.map)"
         ]
       }
     ]


packages/b/buckets.json
```diff
@@ -24,13 +24,13 @@
     "map"
   ],
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/mauriciosantos/Buckets-JS.git",
+    "source": "npm",
+    "target": "buckets-js",
```
MattIPv4, line 28:
```suggestion
    "source": "git",
    "target": "git://github.com/mauriciosantos/Buckets-JS.git",
```
The [git repo has the files needed and tagged correctly](https://github.com/mauriciosantos/Buckets-JS/tree/v1.98.2/dist), so let's avoid complicating the change and stick with git as the source for files.

packages/b/buckets.json
```diff
@@ -24,13 +24,13 @@
...
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "dist",
         "files": [
-          "buckets*.js"
+          "**/*"
```
MattIPv4, line 33:
```suggestion
          "buckets?(.min).js?(.map)"
```
`**/*` is potentially dangerous as it leaves us open to abuse as we'll copy anything published, let's use a more specific glob for the files we need, based on the existing one. [Glob](https://www.digitalocean.com/community/tools/glob?comments=true&glob=buckets%3F%28.min%29.js%3F%28.map%29&matches=false&tests=buckets.js&tests=buckets.min.js&tests=buckets.min.js.map)

PR #330, opened 2020-07-06
Add scheduled package check

klausenbusk:
All packages with autoupdate configured is checked, and if the check
fails a issue is created.

---

Package broken as of today (2020-07-06) is ignored, as creating 688 issues isn't productive.

Ref: https://github.com/cdnjs/packages/issues/302#issuecomment-653818036 and https://github.com/cdnjs/packages/issues/302#issuecomment-653821542


MattIPv4:
My feeling on this is that we should probably hold off on adding such automation until we're in a position where it will be sustainable to actually have these issues created and be able to effectively handle them.

Currently, we have other priorities with the CDN infra itself and we have an existing list of packages (the 688 referenced) that need to be dealt with.

MattIPv4:
Going to close this out as it hasn't moved anywhere for a while, and it seems our current processes are sufficient with folks flagging packages as needing updates when they encounter it.

PR #335, opened 2020-07-08
Update reveal.js w/ git auto-update

hakimel:
We relocated most of our distributed assets in reveal.js 4.0. This PR updates the filemap to cover the new paths.

### DIFF
--- a/packages/r/reveal.js.json
+++ b/packages/r/reveal.js.json
@@ -1,6 +1,6 @@
 {
   "name": "reveal.js",
-  "filename": "dist/reveal.js",
+  "filename": "reveal.js",
   "description": "The HTML Presentation Framework",
   "homepage": "https://revealjs.com/",
   "keywords": [
--- a/packages/r/reveal.js.json
+++ b/packages/r/reveal.js.json
@@ -11,7 +11,7 @@
   ],
   "autoupdate": {
     "source": "git",
-    "target": "git://github.com/hakimel/reveal.js.git",
+    "target": "https://github.com/hakimel/reveal.js.git",
     "fileMap": [
       {
         "basePath": "dist",
--- a/packages/r/reveal.js.json
+++ b/packages/r/reveal.js.json
@@ -1,8 +1,8 @@
 {
   "name": "reveal.js",
-  "filename": "js/reveal.min.js",
+  "filename": "dist/reveal.js",
   "description": "The HTML Presentation Framework",
-  "homepage": "http://lab.hakim.se/reveal-js/",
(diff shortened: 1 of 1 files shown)


packages/r/reveal.js.json
```diff
@@ -16,14 +16,8 @@
...
         "files": [
-          "css/**/*.css",
-          "js/*.js",
-          "lib/**/*.js",
-          "lib/**/*.css",
-          "lib/**/*.eot",
-          "lib/**/*.ttf",
-          "lib/**/*.woff",
-          "plugin/**/*.js"
+          "dist/**",
```
MattIPv4, line 19:
```suggestion
          "dist/**/*.@(js|css|eot|ttf|woff)",
```
Open globstar is dangerous and leaves cdnjs open to abuse, let's specify the file types we need

packages/r/reveal.js.json
```diff
@@ -16,14 +16,8 @@
...
+          "plugin/**"
```
MattIPv4, line 20:
```suggestion
          "plugin/**/*.@(js|css)"
```
As above, open globstar is bad, let's restrict it to what we need :)

hakimel:
Thanks for reviewing. Moving the files in `dist` up to the root makes sense to me. Is this the right way to configure it?

```json
"fileMap": [
  {
    "basePath": "dist",
    "files": [
      "**/*.@(js|css|eot|ttf|woff)",
      "../plugin/**/*.@(js|css)"
    ]
  }
]
```

MattIPv4:
@hakimel To also include plugins, you'd use a second fileMap, so it'd look like this:

```json
"fileMap": [
  {
    "basePath": "dist",
    "files": [
      "**/*.@(js|css|eot|ttf|woff)"
    ]
  },
  {
    "basePath": "",
    "files": [
      "plugin/**/*.@(js|css)"
    ]
  }
]
```

This would result in all the files in `dist` being at the root of the library version in cdnjs, with `plugin` being a subdirectory still.

hakimel:
Thanks @MattIPv4. This should be all set now unless you have any other comments.

packages/r/reveal.js.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "reveal.js",
-  "filename": "js/reveal.min.js",
+  "filename": "dist/reveal.js",
```
MattIPv4, line 3:
```suggestion
  "filename": "reveal.js",
```

As we're using `basePath`, these will be pulled up and there won't be a `dist` subdirectory.

Issue #186, opened 2020-05-29
nerd-fonts is too big


xtuc:
The package `nerf-fonts` includes a lot of files:
```
$ du -sh ajax/libs/nerd-fonts/*
1.1G    ajax/libs/nerd-fonts/0.5.1
1.3G    ajax/libs/nerd-fonts/0.6.0
1.3G    ajax/libs/nerd-fonts/0.6.1
1.7G    ajax/libs/nerd-fonts/0.7.0
3.4G    ajax/libs/nerd-fonts/0.8.0
570M    ajax/libs/nerd-fonts/1.0.0
529M    ajax/libs/nerd-fonts/1.1.0
584M    ajax/libs/nerd-fonts/1.2.0
3.5G    ajax/libs/nerd-fonts/2.0.0
4.0G    ajax/libs/nerd-fonts/2.1.0
```
A version is up to 4GB which is also the same of the commit. GitHub don't accept that.

I'm removing the package for now. No versions were commit to the repo yet.

PR #342, opened 2020-07-12
Add caf w/ npm auto-update

getify:
Library name: caf
Git repository url: https://github.com/getify/caf
npm package name or url (if there is one) : caf
License (List them all if it's multiple): MIT
Official homepage: https://github.com/getify/caf

### DIFF
--- a/packages/c/caf.json
+++ b/packages/c/caf.json
@@ -1,0 +1,30 @@
+{
+  "name": "caf",
+  "description": "Cancelable Async Flows: a wrapper to treat generators as cancelable async functions",
+  "filename": "caf.js",
+  "keywords": [
+    "async",
+    "promises",
+    "generators",
+    "cancelation"
+  ],
+  "author": "Kyle Simpson <getify@gmail.com>",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/getify/caf.git"
+  },
+  "license": "MIT",
+  "homepage": "https://github.com/getify/caf",
+  "autoupdate": {
+    "source": "npm",
+    "target": "caf",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
+        ]
+      }
+    ]
+  }
+}
--- a/packages/c/caf.json
+++ b/packages/c/caf.json
@@ -22,7 +22,7 @@
       {
         "basePath": "dist",
         "files": [
-          "**/*"
+          "**/*.?(js|json|mjs|map)"
         ]
       }
     ]


packages/c/caf.json
```diff
@@ -0,0 +1,30 @@
...
+  "license": "MIT",
+  "homepage": "https://github.com/getify/caf",
+  "autoupdate": {
+    "source": "npm",
+    "target": "caf",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 25:
```suggestion
          "*.js"
```
An open globstar is dangerous and leaves cdnjs open to abuse, let's use a more specific glob for the files needed.

[Glob preview](https://www.digitalocean.com/community/tools/glob?comments=true&glob=dist%2F%2A.js&matches=false&tests=%2F%2F%20Imported%20from%20caf%20NPM%20package&tests=.editorconfig&tests=tests%2Findex.html&tests=dist%2Fabortcontroller-polyfill-only.js&tests=scripts%2Fbuild-core.js&tests=dist%2Fcaf.js&tests=src%2Fcaf.src.js&tests=index.js&tests=scripts%2Fnode-tests.js&tests=tests%2Fqunit.config.js&tests=tests%2Ftests.js&tests=package.json&tests=README.md&tests=src%2Fcopyright-header.txt&tests=LICENSE.txt&tests=.github%2FFUNDING.yml)

getify: The contribution instructions literally list the "**/*" glob as an example. In addition, I randomly looked at 5 or 6 different "recent" library json's, and half of them had that same glob in it. I don't care that much, but being more specific creates more maintenance burden. You also shouldn't list things in the contribution guide that you don't actually want, and you shouldn't say "look at what others are doing recently and do that" if you don't want it.

An open globstar in the package root might be a little dangerous, but it's a dist/ folder... it's clearly for distribution files. I'm not sure what kind of "abuse" you're imagining?

What happens if I later decide to change the structure (or contents) of my dist/ folder? Like, what if I start having sub-directories in it? What if the file names change from .js to .mjs or .cjs? What if I add .map (sourcemap) files?

MattIPv4: > The contribution instructions literally list the "**/*" glob as an example.

Ah, my mistake, that should be updated 👍 

> In addition, I randomly looked at 5 or 6 different "recent" library json's, and half of them had that same glob in it.

I'm very actively trying to avoid merging PRs with `**/*` being used, so recently merged JSON package files should be a good guide.

> I'm not sure what kind of "abuse" you're imagining?

https://github.com/cdnjs/packages/issues/186 would act as a good example of why an open globstar poses potential issues. If we add a package with an open globstar, the project maintainer can then publish anything they want in the target directory and we'll blindly clone it, which can cause major issues on our end.

> Like, what if I start having sub-directories in it?

We try our best to catch packages that are no longer automatically updating and we'll fix the config in them, but if you think this is something you are likely to do soon, then we could add a globstar here, with correct extensions still defined.

> What if the file names change from .js to .mjs or .cjs?

Neither of those extensions are supported by cdnjs, so that'd be a separate issue.

> What if I add .map (sourcemap) files?

Again, if you think this is something you are likely to do soon, we can include that in the glob now (eg. `*.(js|map)`), or we can update the config in a later PR.

getify: > ...a good example of why an open globstar poses potential issues. If we add a package with an open globstar, the project maintainer can then publish anything they want in the target directory and we'll blindly clone it, which can cause major issues on our end.

I understand your concern about not having too many, or too big, of files in your CDN, such that it burdens the servers/services.

But controlling that with glob restrictions seems... ineffective. Of course I could update my package to have 1 million `.js` files in my `dist/` folder, or I could publish an update that has a single `.js` file that is a concat of an entire 1GB node_modules folder.

> Neither of those extensions are supported by cdnjs, so that'd be a separate issue.

You're not supporting people serving up `.mjs` ES module files to browsers? That file extension is, to an extent, required for ES modules to work in Node, so most OSS packages are converting to `.mjs` file extensions for those ES modules. Not supporting it is quite a limitation for modern JS.

In my case, I've written a tool (called [Moduloze](https://github.com/getify/moduloze)) for use in OSS projects that run in both Node and browsers, which takes CJS modules (either .js or .cjs extensions) and converts them to `*.js` (for UMD format) and `*.mjs` (for ESM format), for deployment in `dist/` folders. My plan is to incorporate Moduloze build into CAF quite soon. I already have Moduloze incorporated into my other project, [Monio](https://github.com/getify/monio), which I am planning to submit a PR to add to cdnjs, once this PR is worked out.

The `dist/` folder in that case would likely look like this:

```cmd
dist/esm/
   - caf.mjs
   - abortcontroller-polyfill-only.js
dist/umd/
   - caf.js
   - abortcontroller-polyfill-only.js
```

So, what can we do to get `.mjs` files supported?

MattIPv4: > But controlling that with glob restrictions seems... ineffective. Of course I could update my package to have 1 million .js files in my dist/ folder, or I could publish an update that has a single .js file that is a concat of an entire 1GB node_modules folder.

You sure could. Restricting the glob doesn't prevent abuse happening, but it helps.

> So, what can we do to get .mjs files supported?

@tc80 / @xtuc could we get `cjs` & `mjs` added to the whitelist?

xtuc: > @tc80 / @xtuc could we get cjs & mjs added to the whitelist?
 
yes, sure. I'll let you know once it's live

getify: So... is `**/*.(js|json|mjs|map)` an acceptable glob to use?

MattIPv4: If you know your project is going to have a structure like the above, I'd much prefer something like `@(esm|umd)/*.@(js|json|mjs|map)` (`{@(esm|umd)/,}` if we also need files at the root), but if you're completely unsure then, yeah, the globstar would be the only option.

getify: > If you know ... but if you're completely unsure...

My point is, I neither definitely know (since right now it isn't in that structure, and I'm not sure when it will be) nor am I completely unsure (because my other project has that structure, and so far it seems OK, but it's not on cdnjs yet so I can't be certain yet). I'm in between.

And since I'm in between, I'm trying to gauge the balance that you're wanting to strike between being as strict as possible and being as loose as possible.

* As strict as possible: I could name all the individual files and skip glob wildcards at all

* As loose as possible: the `**/*` globstar

You already pushed back on the latter of those, but I'm trying to figure out where we're supposed to draw the line between that option and the former one at the other extreme.

The tradeoff is maintenance... not only for me (since I have to come back and do another fork/PR here) but for you (since it's not automated and you all still need to look, and discuss if you don't like what I'm doing).

So I'm just asking for some better guidelines on where you want/need that balance line to be drawn. I'm not only asking for *this* project, but to make sure that the next project I want to add, and the four after that (someday), that I don't run into the same mismatch.

MattIPv4: Sounds to me like a restricted globstar is the best bet here 👍

getify: Is there anything further for me to do?

getify: Is this now too old of a fork and needs to be re-based?

MattIPv4: Nope, should be fine. Will review later.

getify:
Thanks, appreciate the guidance!

PR #353, opened 2020-07-22
Update p5.js w/ npm auto-update

akshay-99:
Hi there!
Updating p5.js to also include the `translations/` path. 
The path doesn't exist in the current release's tarball but it will be available in subsequent releases after processing/p5.js#4701 is merged. I hope that's okay :smile:

### DIFF
--- a/packages/p/p5.js.json
+++ b/packages/p/p5.js.json
@@ -27,6 +27,12 @@
         "files": [
           "**/p5*"
         ]
+      },
+      {
+        "basePath": "",
+        "files": [
+          "translations/*/*.json"
+        ]
       }
     ]
   }


packages/p/p5.js.json
```diff
@@ -27,6 +27,12 @@
         "files": [
           "**/p5*"
         ]
+      },
+      {
+        "basePath": "",
+        "files": [
+          "translations/**"
```
MattIPv4, line 34: Open globs are dangerous, please could we make this more specific?
From the PR it sounds like this glob would pick up everything?
```suggestion
          "translations/*/*.json"
```

akshay-99: Thanks! I think this should work. Made the change.

PR #355, opened 2020-07-23
Standardizing license field

tc80:
No more `licenses` field, just `license`. [For reference](https://docs.npmjs.com/files/package.json#license).

Will update the CI to enforce `license` in another PR, as well as find any stragglers that haven't been fixed....

```
find -name '*.json' | xargs -n 1 -P 10 -I {} bash -c 'jq "if .license.name then . + {license: .license.name} else . end" "{}" > "{}.tmp" && mv "{}.tmp" "{}"'

find -name '*.json' | xargs -n 1 -P 10 -I {} bash -c 'jq "if .license.type then . + {license: .license.type} else . end" "{}" > "{}.tmp" && mv "{}.tmp" "{}"'

find -name '*.json' | xargs -n 1 -P 10 -I {} bash -c 'jq "if .licenses then . + {license: (\"(\"+(.licenses | join(\" OR \"))+\")\")} | del(.licenses) else . end" "{}" > "{}.tmp" && mv "{}.tmp" "{}"'
```

Credit to @klausenbusk . I tweaked your command from before 😉

### DIFF
--- a/packages/9/960gs.json
+++ b/packages/9/960gs.json
@@ -12,8 +12,5 @@
     "type": "git",
     "url": "https://github.com/nathansmith/960-Grid-System/"
   },
-  "licenses": [
-    "GPL",
-    "MIT"
-  ]
+  "license": "(GPL OR MIT)"
 }
--- a/packages/a/a-happy-tyler.json
+++ b/packages/a/a-happy-tyler.json
@@ -1,15 +1,14 @@
 {
   "name": "a-happy-tyler",
   "description": "Tyler is happy. Be like Tyler.",
-  "keywords": [],
-  "author": {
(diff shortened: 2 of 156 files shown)


tc80:
Every `license` field is now a string.

Need to fix `licenses` stragglers...

tc80:
Update:

All libraries now contain `license` and it is a `string`. The following libraries are missing the property, but it is optional so this is OK:

```
[1/1140.json a/animations.json b/boexfi.json c/chrome-frame.json e/echarts-gl.json j/jcalculator.json l/luminateExtend.json m/mojio-js.json p/probtn.json]
```

Also, there are now no packages that contain `licenses` 👍

PR #356, opened 2020-07-24
Standardize author field

tc80:
Author is optional. There are currently 3464 packages with it and 450 without.

I standardized author to only contain the string fields:

```
email (679)
name (3464)
phone (1)
twitter (8)
url (889)
```

In parentheses is the count of packages with this property. I refactored several `url` equivalent properties to `url`, as well as removed several unnecessary properties that existed in only one or two packages (ex. `company`).

Note that `author` cannot be just a string anymore (changed `{"author": "<name>"}` -> `{"author": {"name": "<name>"}}`):
```
find -name '*.json' | xargs -n 1 -P 10 -I {} bash -c 'jq "if (.author | type == \"string\") then . + {author: {name: .author}} else . end" "{}" > "{}.tmp" && mv "{}.tmp" "{}"'
```
Some of the `name` fields also contained emails, but this is not a major issue.

Example command for changing `homepage` -> `url`:
```
find -name '*.json' | xargs -n 1 -P 10 -I {} bash -c 'jq "if .author.homepage then .author += {url: .author.homepage} | del(.author.homepage) else . end" "{}" > "{}.tmp" && mv "{}.tmp" "{}"'
```


tc80:
Removed `author` field and put into the `authors` array of `author` objects. Based on #361 .

Command:
```
find -name '*.json' | xargs -n 1 -P 10 -I {} bash -c 'jq "if .author then .authors = [.author] | del(.author) else . end" "{}" > "{}.tmp" && mv "{}.tmp" "{}"'
```

MattIPv4:
Two quick thoughts:

Entries like `"name": "Jonathan Neal <jonathantneal@hotmail.com> (https://jonneal.dev/)"` should really be parsed into name, email & url. The use of <> & () is somewhat standard I think.

Should this be considered a breaking change for the API?

(And yeah, the API logic will need updating as will the API docs in cdnjs/static-website)

tc80:
yep sounds good, can do that!

This won't break the API if it relies on the package.min.json generation. If it reads from the repo directly it might!

tc80:
Parsed `name <email> (url)`:

```
find -name '*.json' | while read l; do cat $l | jq 'if .authors then (.authors[] |= if .name | test("\\s*(.*[^\\s])\\s*<(.+)>\\s*\\((.+)\\)") then (.name | capture("\\s*(?<n>.*[^\\s])\\s*<(?<e>.+)>\\s*\\((?<u>.+)\\)") | {name: .n, email: .e, url: .u}) else . end) else . end' > $l.tmp && mv $l.tmp $l && echo $l; done
```

Will parse just `name <email>` and `name (url)` next.

tc80:
Parsed `name <email>`:

```
find -name '*.json' | while read l; do cat $l | jq 'if .authors then (.authors[] |= if .name | test("\\s*(.*[^\\s])\\s*<(.+)>\\s*") then (.name | capture("\\s*(?<n>.*[^\\s])\\s*<(?<e>.+)>\\s*") | {name: .n, email: .e}) else . end) else . end' > $l.tmp && mv $l.tmp $l && echo $l; done
```

tc80:
Parsed `name (url)`.
```
find -name '*.json' | while read l; do cat $l | jq 'if .authors then (.authors[] |= if .name | test("\\s*(.*[^\\s])\\s*\\((.+)\\)") then (.name | capture("\\s*(?<n>.*[^\\s])\\s*\\((?<u>.+)\\)") | {name: .n, url: .u}) else . end) else . end' > $l.tmp && mv $l.tmp $l && echo $l; done
```

I went through the changed files manually and fixed any odd JSON files. Lots of people put their handles in parentheses, and some put `email` by accident.

tc80:
> Entries like "name": "Jonathan Neal <jonathantneal@hotmail.com> (https://jonneal.dev/)" should really be parsed into name, email & url. The use of <> & () is somewhat standard I think.

Should be all set @MattIPv4. I can add some ci checker logic to ensure any names containing `()` or `<>` have warnings.

MattIPv4:
> This won't break the API if it relies on the package.min.json generation. If it reads from the repo directly it might!

How so? We'll now be returning `authors` in responses, not `author`? Should we not consider that a breaking change for users of the API? Or are we okay just doing it, as its an optional field currently?

MattIPv4:
In a similar line of thought, this will also change how our Algolia index behaves? Presumably that'll now need to index authors, not author?

tc80:
ah sorry @MattIPv4 now I get what you mean.

When we convert the human readable JSON to the package.json (containing `version`) (and will also be stored elsewhere soon), maybe we just parse `authors` into an additional field `author`. We can just have a string with comma (or some alternative) separated author names

For example for:

```
"authors": [
    {
      "name": "Andrey Tarantsov",
      "email": "andrey@tarantsov.com"
    },
    {
      "name": "David Kelso",
      "email": "david@kelso.id.au"
    }
  ]
```

We can add an additional field (that is stored in KV along with `version`):
```
"author": "Andrey Tarantsov,David Kelso
```

This will make sure the API keeps the `author` field while adding the new `authors` field, and we can enforce that new PRs only contain `authors`

MattIPv4:
Yeah that makes sense to map it back to the `author` field. We should do that at the same point in the code that we're going to add the `filename` fix, so the new `author` field  exists in Algolia & packages.min.json.

When you convert it to `author`, you could even map `url` & `email` back in the standard format with () and <>.

tc80:
yup. If someone has a name like `Matt (MattIPv4)`, should I just tack on another `()` for url? I guess it is their problem, and if you parse from right to left you should always get the url/email

MattIPv4:
> should I just tack on another () for url?

Yeah, no need to add any logic that makes it more complicated :)

tc80:
Manually merged with master from command-line 😢 . Had to ssh into linux because my mac cannot handle the case sensitive paths.

Checked my changes are all valid JSON with 
```
find -name '*.json' | while read l; do cat $l | jq '.' > $l.tmp && mv $l.tmp $l && echo $l; done
```

Will remove `phone` and remove/refactor the `twitter` properties manually.

tc80:
Fixed all `twitter` properties. Found them with:
```
find -name '*.json' | while read l; do echo "$(cat $l | jq 'if .authors then (.authors | .. | if .twitter? then "HIT: "+.twitter else "MISS" end) else "MISS" end') - $l"; done
```

tc80:
`phone` removed

PR #360, opened 2020-07-24
Standardize filename field

tc80:
[mavo](https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/mavo) has `main` instead of `filename`. This is the only package with top-level property `main`.

### DIFF
--- a/packages/m/mavo.json
+++ b/packages/m/mavo.json
@@ -1,7 +1,7 @@
 {
   "name": "mavo",
   "description": "Create web applications with HTML and CSS",
-  "main": "mavo.js",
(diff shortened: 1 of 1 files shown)


packages/m/mavo.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mavo",
   "description": "Create web applications with HTML and CSS",
-  "main": "mavo.js",
+  "filename": "mavo.js",
```
MattIPv4, line 4:
```suggestion
  "filename": "mavo.min.js",
```
Let's recommend the minified version while we're here

PR #361, opened 2020-07-24
Standardizing authors field

tc80:
We are currently not using `authors`, but 89 packages defines them. I think we should remove the `author` field in favor of the `authors` field, and make the `authors` field an array of `author` objects.

Ran:
```
find -name '*.json' | xargs -n 1 -P 10 -I {} bash -c 'jq "if .authors then .authors = [.authors[] | {name: .}] else . end" "{}" > "{}.tmp" && mv "{}.tmp" "{}"'
```
Then cleaned up the stragglers.

### DIFF
--- a/packages/z/zrender.json
+++ b/packages/z/zrender.json
@@ -11,16 +11,12 @@
   "homepage": "https://ecomfe.github.io/zrender/",
   "authors": [
     {
-      "name": {
-        "name": "Kenner",
-        "email": "kener.linfeng@gmail.com"
-      }
+      "name": "Kenner",
+      "email": "kener.linfeng@gmail.com"
     },
     {
-      "name": {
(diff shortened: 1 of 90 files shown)


packages/z/zrender.json
```diff
@@ -11,16 +11,22 @@
...
+        "url": "https://github.com/pissang"
+      }
     },
     {
-      "name": "Shuang Su",
-      "url": "https://github.com/100pah"
+      "name": {
+        "name": "Shuang Su",
+        "url": "https://github.com/100pah"
+      }
```
xtuc, line 29: This doesn't look right

tc80: bad sven, I wasn't ready!!! 😆

PR #378, opened 2020-07-30
Add polyavatar w/ npm auto-update

programad:


packages/p/polyavatar.json
```diff
@@ -0,0 +1,36 @@
...
+    "url": "git://github.com/programad/polyavatar.git"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "polyavatar",
+    "fileMap": [
+      {
+        "basePath": "src",
+        "files": [
+          "**/*"
```
MattIPv4, line 17:
```suggestion
          "*.js"
```

Using open globstars is dangerous and leaves cdnjs open to potential abuse -- let's use a restricted wildcard that just gets the files we need. [glob tool](https://www.digitalocean.com/community/tools/glob?comments=true&glob=src%2F%2A.js&matches=false&tests=%2F%2F%20Imported%20from%20polyavatar%20NPM%20package&tests=LICENSE&tests=src%2Fpolyavatar.js&tests=src%2Fpolyavatar.min.js&tests=package.json&tests=README.md)

packages/p/polyavatar.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "polyavatar.js",
```
MattIPv4, line 2:
```suggestion
  "name": "polyavatar",
```
`name` should match the JSON file name

programad: Agree. Thank you.

PR #383, opened 2020-08-06
Add lit-grid-layout w/ npm auto-update

zsarnett:
Adding Lit-grid-layout w/ npm auto-update

### DIFF
--- a/packages/l/lit-grid-layout.json
+++ b/packages/l/lit-grid-layout.json
@@ -1,6 +1,10 @@
 {
   "name": "lit-grid-layout",
-  "author": "Zack Arnett",
+  "authors": [
+    {
+      "name": "Zack Arnett"
+    }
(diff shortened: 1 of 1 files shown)


packages/l/lit-grid-layout.json
```diff
@@ -0,0 +1,35 @@
...
+    "url": "git+https://github.com/zsarnett/Lit-Grid-Layout.git"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "lit-grid-layout",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*.js"
```
MattIPv4, line 30:
```suggestion
          "**/*@(.js|.ts)?(.map)"
```
cdnjs support map files & typings, so we should include those.

PR #403, opened 2020-08-19
Add tonic w/ npm autoupdate

### LINKED ISSUE #402: Add: @optoolco/tonic
**Library name:** @optoolco/tonic

**Library license (include link):** MIT ( https://github.com/optoolco/tonic/blob/master/LICENSE.txt )

**Library source (NPM package or Git repo):** https://github.com/optoolco/tonic

One of our users requested cdnjs support https://github.com/optoolco/tonic/issues/57

Raynos:
Adding tonic using npm auto-update from optoolco/tonic.

Resolves #402.

### DIFF
--- a/packages/t/tonic.json
+++ b/packages/t/tonic.json
@@ -1,6 +1,5 @@
 {
   "name": "tonic",
-  "filename": "tonic.js",
   "description": "A composable component inspired by React.",
   "keywords": [
     "tonic",
@@ -9,10 +8,22 @@
     "components",
     "html"
   ],
+  "license": "MIT",
   "homepage": "https://tonic.technology/",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/optoolco/tonic.git"
+  },
+  "authors": [
+    {
+      "name": "Raynos",
+      "email": "raynos2@gmail.com",
+      "url": "https://github.com/optoolco/tonic/issues"
+    }
+  ],
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/optoolco/tonic.git",
+    "source": "npm",
+    "target": "@optoolco/tonic",
     "fileMap": [
       {
         "basePath": "",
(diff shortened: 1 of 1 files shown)


packages/t/tonic.json
```diff
@@ -0,0 +1,37 @@
...
+  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "@optoolco/tonic",
+    "fileMap": [
+      {
+        "basePath": "",
+        "files": [
+          "index.js"
+        ]
```
MattIPv4, line 32:
```suggestion
        "basePath": "dist",
        "files": [
          "*.js"
        ]
```

It looks like you provide a pre-bundled distributable version. We should use this instead as a Node version won't work on the web.

MattIPv4 (changes requested):

packages/t/tonic.json
Raynos: The pre-bundled variant is minified by default.

The Node version does happen to work on the web because the exports is a conditional

```
if (typeof module === 'object') module.exports = Tonic
```

Raynos:
Our package has two files

 - `index.js` ; raw source code, not minified, works in a `<script>` tag due to `if (typeof module === 'object') ...`
 - `dist/tonic.min.js` ; minified version, same code, same `if (typeof module === 'object') ...` gaurd.

Tonic works in a `<script>` tag because it contains `class Tonic {}` which becomes a global variable ( It also exposes a second `TonicRaw` class as a global variable too ).

Would you recomend that cdnjs host the `index.js` or the `tonic.min.js` or both ?

Raynos:
I applied your suggestion by adding both `index.js` and `dist` to the `fileMap`.

Raynos:
I've addressed feedback on this pull request. Could you take a look again ? Thanks for the review~

MattIPv4 (approved):
Lgtm -- thanks!

Raynos:
Thanks matt!

PR #409, opened 2020-08-20
Update jsSHA w/ git auto-update

### LINKED ISSUE #401: Update: jsSHA
**Library name:** [jsSHA](https://cdnjs.com/libraries/jsSHA)

**Library JSON file link:** https://github.com/cdnjs/packages/blob/master/packages/j/jsSHA.json

**Library source (NPM package or Git repo):** https://github.com/Caligatio/jsSHA

**Files to include:** Existing files

The latest version cdnjs is 2.4.2 but there have been [several releases since then](https://github.com/Caligatio/jsSHA/releases), with the current version being 3.1.2.

Thanks

MattIPv4:
Fixes #401

### DIFF
--- a/packages/j/jsSHA.json
+++ b/packages/j/jsSHA.json
@@ -27,9 +27,9 @@
     "target": "git://github.com/Caligatio/jsSHA.git",
     "fileMap": [
       {
-        "basePath": "src",
+        "basePath": "dist",
         "files": [
-          "*.js"
+          "*.@(@(js|mjs)?(.map)|d.ts)"
         ]
       }
     ]


DanAtkinson:
Hi Matt,

Can I confirm that, when this gets merged into master, it will result in the missing versions automatically being added?

Thanks, Dan

MattIPv4:
I can't guarantee that it will backfill old missing versions (it might pick up some, unsure off the top of my head), but it will most definitely pick up any new versions published after this is merged, that match the new config.

DanAtkinson:
Thanks. If it doesn't pick up new version (or even latest), this seems like a hole in the functionality that _should_ be resolved - either by allowing manual linking to older releases, or by creating a mechanism that allows them to be discovered at a later date.

I don't pretend to understand the jobs that run behind the scenes to update these but, in the context of jsSHA, it appears that the structure of the repository changed after the last 'good' version was released.

When something in a package changes and the files retrieved are not as expected (e.g a file/folder doesn't exist) is there a mechanism to log or flag this for investigation?

PR #412, opened 2020-08-21
Add map files to list of retrieved files

DanAtkinson:
The `.map` files are missing from the bootstrap 4.5.1 and 4.5.2 releases and causes console warnings as a result. However, looking at the diffs for the last few versions, I can't see a reason why they would have gone missing.

This change _should_ ensure they're being pulled in for future releases.


MattIPv4 (approved):
Lgtm -- thank you :)

PR #431, opened 2020-09-08
Update sip.js w/ npm auto-update

james-criscuolo:
We've updated our dist folder from `dist` to `lib` since moving to typescript, and the initial file to pull is now `index.js` there instead of `sip.min.js`

### DIFF
--- a/packages/s/sip.js.json
+++ b/packages/s/sip.js.json
@@ -15,13 +15,13 @@
     "type": "git",
     "url": "git+https://github.com/onsip/SIP.js.git"
   },
-  "filename": "index.js",
+  "filename": "sip.min.js",
   "autoupdate": {
-    "source": "npm",
-    "target": "sip.js",
+    "source": "git",
+    "target": "git://github.com/onsip/SIP.js.git",
     "fileMap": [
       {
-        "basePath": "lib",
+        "basePath": "dist",
         "files": [
           "*.js"
         ]
--- a/packages/s/sip.js.json
+++ b/packages/s/sip.js.json
@@ -15,13 +15,13 @@
     "type": "git",
     "url": "git+https://github.com/onsip/SIP.js.git"
   },
-  "filename": "sip.min.js",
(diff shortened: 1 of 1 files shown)


MattIPv4 (changes requested):
`lib/index.js` is a Node.js file that uses require. This will not work in the context of a browser. The old `dist/sip.min.js` file was a prebundled version, which is what's needed to work in a browser.

james-criscuolo:
Sorry for this mistake. We stopped publishing our bundles via NPM, and now only have them in git. It appears I can switch to git based in my json file, but I don't see anything in the documentation (or other libraries using git) of not using the master branch. Our bundles live on the release branch in the `dist` folder, here: https://github.com/onsip/SIP.js/tree/release

Is it possible to declare a branch in the json file?

MattIPv4:
@tc80 do we check out the tag directly when getting versions from git? If so, then a branch setting isn't needed as a tag is always tied to a specific branch/commit

james-criscuolo:
Our tags do point to that branch, so we would be ok there too, assuming it just takes the latest tag?

MattIPv4:
Yeah, we work based on tags for git auto-update. If you update this PR to use git auto-update, the CI will test if its working as expected or not.

tc80:
> @tc80 do we check out the tag directly when getting versions from git? If so, then a branch setting isn't needed as a tag is always tied to a specific branch/commit

```// ForceCheckout force checkouts a particular tag.
func ForceCheckout(ctx context.Context, gitpath, tag string) {
	args := []string{"checkout", tag, "-f"}

	cmd := exec.Command("git", args...)
	cmd.Dir = gitpath
	util.Debugf(ctx, "run %s\n", cmd)
	util.CheckCmd(cmd.CombinedOutput())
}```

Looks to me like we check out the tag directly. Should be ok! 👍

MattIPv4 (approved):
Lgtm - thank you! :)

PR #438, opened 2020-09-13
Add vuejs-storage w/ npm auto-update

maple3142:
Adding vuejs-storage

### DIFF
--- a/packages/v/vuejs-storage.json
+++ b/packages/v/vuejs-storage.json
@@ -1,0 +1,36 @@
+{
+  "name": "vuejs-storage",
+  "description": "Vue.js and Vuex plugin to persistence data with localStorage/sessionStorage",
+  "keywords": [
+    "vue",
+    "vuex",
+    "localStorage",
+    "sessionStorage",
+    "persistence"
+  ],
+  "license": "MIT",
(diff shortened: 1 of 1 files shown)


packages/v/vuejs-storage.json
```diff
@@ -0,0 +1,36 @@
...
+    "url": "https://github.com/maple3142/vuejs-storage.git"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "vuejs-storage",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "*.js"
```
MattIPv4, line 25:
```suggestion
          "*.js?(.map)"
```
We can also bring in the map files :)

packages/v/vuejs-storage.json
```diff
@@ -0,0 +1,36 @@
...
+        ]
+      }
```
MattIPv4, line 27:
```suggestion
      },
      {
        "basePath": "types",
        "files": [
          "*.d.ts"
        ]
      }
```
We can also bring in typings files :)

MattIPv4 (changes requested):

MattIPv4 (approved):
Lgtm, thanks!

PR #447, opened 2020-09-21
Update nipplejs w/ npm auto-update

yoannmoinet:
We've updated our way of delivering the lib.
We don't ship built files anymore, they are built in the prepack script instead.

This PR reflects this change.


MattIPv4 (approved):
Lgtm!

PR #452, opened 2020-09-29
Add easy-toggle-state w/ git auto-update

Twikito:


MattIPv4 (changes requested):
:wave: So that we can suggest a default file for folks to use with this library, please could you add the `filename` property? I imagine the default file to suggest would be `easy-toggle-state.min.js`?

Twikito:
Actually no, it's `easy-toggle-state.es6.min.js` :)

MattIPv4 (approved):
Lgtm, thanks!

PR #476, opened 2020-10-26
Update echarts w/ git auto-update

plainheart:
Hi, the cdnjs team,

This PR intends to update the description and keywords of `echarts` library.
Also, according to the 5.x version, I added two new directories `theme` and `i18n`.

Thank you.

### DIFF
--- a/packages/e/echarts.json
+++ b/packages/e/echarts.json
@@ -1,11 +1,12 @@
 {
   "name": "echarts",
-  "description": "A powerful charting and visualization library for browser",
+  "description": "A powerful, interactive charting and data visualization library for browser",
   "filename": "echarts.min.js",
   "keywords": [
-    "baidu",
+    "apache",
     "echarts",
     "canvas",
+    "svg",
     "data visualization"
   ],
   "autoupdate": {
@@ -14,6 +15,12 @@
     "fileMap": [
       {
         "basePath": "dist",
+        "files": [
+          "**/*"
+        ]
+      },
+      {
+        "basePath": "i18n",
         "files": [
           "**/*"
         ]
--- a/packages/e/echarts.json
+++ b/packages/e/echarts.json
@@ -16,13 +16,13 @@
       {
         "basePath": "dist",
         "files": [
-          "**/*"
+          "**/*.js?(.map)"
         ]
       },
       {
         "basePath": "i18n",
         "files": [
-          "**/*"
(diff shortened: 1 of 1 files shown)


packages/e/echarts.json
```diff
@@ -17,6 +18,18 @@
         "files": [
           "**/*"
         ]
+      },
+      {
+        "basePath": "i18n",
+        "files": [
+          "**/*"
```
MattIPv4, line 25:
```suggestion
          "*.js"
```

packages/e/echarts.json
```diff
@@ -17,6 +18,18 @@
         "files": [
           "**/*"
```
MattIPv4, line 19:
```suggestion
          "**/*.js?(.map)"
```

MattIPv4 (changes requested):
Open wildcards are dangerous, let's at least restrict them with file extensions that are expected

plainheart:
Thanks for your kind reminder, I got it.

packages/e/echarts.json
```diff
@@ -15,7 +16,19 @@
...
+        "files": [
+          "*.js"
+        ]
+      },
+      {
+        "basePath": "theme",
+        "files": [
+          "**/*.js"
         ]
       }
```
MattIPv4, line 33: These are pulling the files from `i18n/*.js` to `*.js` without the subdirectory -- do you want to preserve the subdirectory?

plainheart: Not necessary. `i18n` has commonly no subdirectory. Thanks.

MattIPv4: The same goes for theme, all the files from `theme` are being pulled up into the root directory -- I'm concerned there might be file collision and it'll be confusing for folks?

plainheart: Could we make two directories `i18n` and `theme` in the root directory? It may be confusing indeed if they are all in the root directory.

MattIPv4:
```suggestion
      {
        "basePath": "",
        "files": [
          "i18n/*.js",
          "theme/**/*.js"
        ]
      }
```

plainheart: Thank you for patient help. Is there anything else that needs to be concerned and corrected?

MattIPv4 (approved):
Lgtm

PR #486, opened 2020-11-01
Add openplayerjs w/ git auto-update

rafa8626:


MattIPv4 (changes requested):
Please indent the file with two spaces, not four, and ensure there is a space between each key and its value.

rafa8626:
@MattIPv4 changes applied; let me know if anything else is needed

rafa8626 (commented):
Changes applied

packages/o/openplayerjs.json
```diff
@@ -0,0 +1,54 @@
...
+        "basePath": "dist",
+        "files": ["**/*"]
+      },
+      {
+        "basePath": "types",
+        "files": ["**/*"]
+      },
+      {
+        "basePath": "docs",
+        "files": ["**/*"]
```
MattIPv4, line 44:
```suggestion
```
I don't think the CDN needs to serve the raw docs files.

packages/o/openplayerjs.json
```diff
@@ -0,0 +1,54 @@
...
+    "source": "git",
+    "target": "https://github.com/openplayerjs/openplayerjs.git",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": ["**/*"]
+      },
+      {
+        "basePath": "types",
+        "files": ["**/*"]
```
MattIPv4, line 40:
```suggestion
        "basePath": "",
        "files": [
          "types/**/*.d.ts"
        ]
```
Let's keep `types` in their subdirectory and restrict the wildcard so that it's not open and dangerous.

packages/o/openplayerjs.json
```diff
@@ -0,0 +1,54 @@
...
+    "type": "git"
+  },
+  "license": "MIT",
+  "autoupdate": {
+    "source": "git",
+    "target": "https://github.com/openplayerjs/openplayerjs.git",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": ["**/*"]
```
MattIPv4, line 36:
```suggestion
        "files": [
          "**/*.@(js|css)"
        ]
```
Open wildcards are dangerous, let's restrict it to just the files we need.

MattIPv4 (changes requested):

rafa8626:
@MattIPv4 I applied all the suggestions you indicated; let me know if you need anything else

MattIPv4 (approved):
Lgtm

PR #488, opened 2020-11-02
Add numl w/ auto-update

tenphi:


MattIPv4 (changes requested):
As per the CI, you need to add an auto-update config for the package. See https://github.com/cdnjs/packages/blob/master/CONTRIBUTING.md#configuring-library-auto-update

tenphi:
@MattIPv4 Sorry. I missed that. Fixed.

packages/n/numl.json
```diff
@@ -0,0 +1,39 @@
...
+  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "numl",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "*.js",
+          "dev/*.js"
```
MattIPv4, line 34: This directory doesn't exist?

tenphi: Yep, it doesn't exist. But it will be there after the next version. If it's a problem, we can wait for a day or two until I publish a new version.

tenphi: @MattIPv4  The directory now exists. Sorry for the inconvenience.

MattIPv4 (approved):
Lgtm

PR #498, opened 2020-11-11
Add devextreme-aspnet-data w/ npm auto-update

OlegKipchatov:


OlegKipchatov:
Hello @MattIPv4,
The check says that the latest version of the package does not contain the file specified in 'filename', but I definitely see it in the package. Would you please describe the issue in greater detail?

packages/d/devextreme-aspnet-data.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "devextreme-aspnet-data",
+  "filename": "js/dx.aspnet.data.js",
```
MattIPv4, line 3:
```suggestion
  "filename": "dx.aspnet.data.js",
```
You're using basePath `js`, so the files are pulled up from that subdirectory into the root :)

MattIPv4 (changes requested):

packages/d/devextreme-aspnet-data.json
OlegKipchatov: Thank you very much :)

MattIPv4 (approved):
Lgtm

PR #499, opened 2020-11-11
Add devexpress-gantt w/ npm auto-update

OlegKipchatov:


packages/d/devexpress-gantt.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "devexpress-gantt",
+  "filename": "dist/dx-gantt.min.js",
```
MattIPv4, line 3:
```suggestion
  "filename": "dx-gantt.min.js",
```
`dist` is the base path, so files are pulled from that into the root.

MattIPv4 (changes requested):

MattIPv4 (approved):
Lgtm

PR #524, opened 2020-11-30
Add mitojs m/ npm auto-update

cjinhuo:


packages/m/mitojs.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@zyf2e/mitojs",
```
MattIPv4, line 2:
```suggestion
  "name": "mitojs",
```
This should match the filename, we only allow alphanum, periods, underscore & hyphens in package names

packages/m/mitojs.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "filename": "index.min.js",
+  "keywords": ["monitor", "sdk", "error", "broswer"],
+  "license": "MIT",
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/clouDr-f2e/mitojs.git",
+    "fileMap": [
+      {
+        "basePath": "dist",
```
MattIPv4, line 16: There is no `dist` directory in your latest tagged version: https://github.com/clouDr-f2e/mitojs/tree/1.1.8

packages/m/mitojs.json
```diff
@@ -0,0 +1,26 @@
...
+        "files": ["**/*"]
```
MattIPv4, line 17: Open wildcards are dangerous. Please restrict this to an exact pattern or at least limit the extensions.

Eg. `**/*.js` or just `*.js`

MattIPv4 (changes requested):

MattIPv4:
No progress, closing.

PR #571, opened 2021-01-02
Add cript.js w/ git auto-update

DablioZe:


MattIPv4 (changes requested):
The latest tagged version (https://github.com/DablioZe/apicriptjs/tree/1.1.3) does not have the lib folder that your auto-update config here is expecting. It looks like your default branch does now (https://github.com/DablioZe/apicriptjs), so you likely just need to publish a new tag.

MattIPv4:
No progress, closing.

PR #611, opened 2021-02-10
Update hyperapp w/ npm auto-update

jorgebucaran:
Hey! So, we changed how we publish Hyperapp. Now we just deploy an `index.js` that describes an ES module. 

I was also wondering if cdnjs still generates a minified version for us automatically. That would be nice.

### DIFF
--- a/packages/h/hyperapp.json
+++ b/packages/h/hyperapp.json
@@ -1,29 +1,29 @@
 {
   "name": "hyperapp",
-  "filename": "hyperapp.js",
-  "description": "1kb JavaScript library for building modern UI applications",
+  "filename": "index.js",
+  "description": "The tiny framework for building hypertext applications",
   "license": "MIT",
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/hyperapp/hyperapp.git"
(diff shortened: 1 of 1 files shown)


MattIPv4:
👋 cdnjs will minify JS files, but we do not perform any bundling, so if you're only shipping an es module, it will only work in modern browsers that support es modules natively -- I strongly recommend shipping pre-bundled cjs/umd versions alongside your es module for web usage.

MattIPv4:
Let me know if you'd still like to ship this with es module support only and I'll be happy to approve/merge :)

jorgebucaran:
> cdnjs will minify JS files

Amazing, that's great! 🎯

> it will only work in modern browsers that support es modules natively 

No problem! 🙆‍♂️  

The upcoming Hyperapp 2 will only support modern browsers. We provide the source, though, so it's possible to compile and polyfill for IE11 and even IE10.

Do you mind reviewing my PR? I don't know if my changes here are valid as I'm not familiar with cdnjs internals.

MattIPv4 (approved):
Lgtm

PR #613, opened 2021-02-11
Add valigators w/ npm auto-update

wkelly1:


packages/v/valigators.json
```diff
@@ -0,0 +1,26 @@
...
+    "type": "git",
+    "url": "https://github.com/wkelly1/Valigators.git"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "valigators",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": ["**/*"]
```
MattIPv4, line 22: Please avoid using open globstar patterns, as they are dangerous and can lead to cdnjs cloning excess files that we don't need.

You likely just need something like `*.js` here.

MattIPv4 (changes requested):

packages/v/valigators.json
wkelly1: Ok thanks, I've updated `"files"` to just the path to the bundle file.

MattIPv4 (approved):
Lgtm

PR #614, opened 2021-02-11
Add diff-match-patch w/ npm auto-update

christiansk:
The version of this library currently in the CDN dates back to 2012. Since then the repo was moved from GoogleCode to GitHub, where a much more recent version is available.

Apologies if I messed anything up in this update, I went by example from recent packages as recommended by the contributing guidelines.


christiansk:
I realize they don't tag the releases in the originating repo for this library. Apologies for setting this up incorrectly.

Current version number is 1.5.0, how can I manually set this so we at least get the latest version in the CDN?

MattIPv4:
👋 Please don't remove the old JSON config file. Old versions on cdnjs will remain under that name, so the JSON file needs to also remain.

Libraries on cdnjs must now have a way to be automatically updated, either via git tags, or a npm package: https://github.com/cdnjs/packages/blob/master/CONTRIBUTING.md#configuring-library-auto-update

christiansk:
Ok, thanks! I'll try to link it to the npm version then.

christiansk:
Ok, I _think_ I got it this time around. Please let me know if this needs any additional tweaking.

MattIPv4 (approved):
lgtm, ty

PR #617, opened 2021-02-16
Update jclic.js w/ npm auto-update

frncesc:
Since the 'dist' folder is no longer included in the GitHub repository,
auto-update source must be changed to 'npm'.

Also, only '*.js' files are now included.


MattIPv4 (approved):
Lgtm

PR #621, opened 2021-02-18
Update fullPage.js w/ git auto-update

alvarotrigo:
https://github.com/alvarotrigo/fullPage.js/blob/master/package.json

### DIFF
--- a/packages/f/fullPage.js.json
+++ b/packages/f/fullPage.js.json
@@ -1,16 +1,27 @@
 {
-  "name": "fullPage.js",
+  "name": "fullpage.js",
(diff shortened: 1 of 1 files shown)


packages/f/fullPage.js.json
```diff
@@ -1,16 +1,27 @@
 {
-  "name": "fullPage.js",
+  "name": "fullpage.js",
```
MattIPv4, line 2:
```suggestion
  "name": "fullPage.js",
```
Must match file name

MattIPv4 (changes requested):

MattIPv4 (approved):
Lgtm

PR #633, opened 2021-03-01
Mark fingerprintjs and fingerprintjs2 as renamed

Finesse:
I'm a FingerpringJS maintainer. The libraries have been renamed to `@fingerprintjs/fingerprintjs` and I try to convince users to switch to the new name. You can open any old link in the configuration files to check that they all redirect to the new repository name.

### DIFF
--- a/packages/f/fingerprintjs.json
+++ b/packages/f/fingerprintjs.json
@@ -1,7 +1,7 @@
 {
   "name": "fingerprintjs",
   "filename": "fingerprint.min.js",
-  "description": "Fast browser fingerprint library. By default uses Murmur hashing and returns a 32bit integer number.",
+  "description": "The library has been renamed to @fingeprintjs/fingerprintjs. Check the link below to get updates.",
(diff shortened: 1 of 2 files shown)


Finesse:
The CI fails because the configuration doesn't support the current library file structure. I don't add support of the new file structure because the library has been renamed, therefore `fingerprintjs2` must provide only the old files. All I want to do is to change the text at https://cdnjs.com/libraries/fingerprintjs2. Is it possible? Alternatively, the following solutions will work for us too:

- set up the configuration to take files only from [a specific git branch](https://github.com/fingerprintjs/fingerprintjs/tree/v2)
- redirect from https://cdnjs.com/libraries/fingerprintjs2 to https://cdnjs.com/libraries/@fingerprintjs/fingerprintjs
- remove https://cdnjs.com/libraries/fingerprintjs2 (leaving the distributive files for backward compatibility)

MattIPv4 (approved):
Lgtm

Finesse:
Thank you!

PR #664, opened 2021-03-20
Update two.js w/ git auto-update

### LINKED ISSUE #663: Update: two.js
**Library name:** two.js

**Library JSON file link:** https://github.com/cdnjs/packages/blob/master/packages/t/two.js.json

**Library source (NPM package or Git repo):** https://github.com/jonobr1/two.js/

**Files to include:** two.module.js

It looks like the regex in the package file for two.js explicitly excludes two.module.js. I looked in the commit history but couldn't find why.

MattIPv4:
Resolves #663

### DIFF
--- a/packages/t/two.js.json
+++ b/packages/t/two.js.json
@@ -21,7 +21,7 @@
       {
         "basePath": "build",
         "files": [
-          "two!(.module)*.js"
+          "*.js"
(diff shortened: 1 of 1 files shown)


MattIPv4 (commented):
lgtm

PR #671, opened 2021-03-25
Add twind w/ npm auto-update

sastan:
Adding twind using npm auto-update .

### DIFF
--- a/packages/t/twind.json
+++ b/packages/t/twind.json
@@ -1,0 +1,31 @@
+{
+  "name": "twind",
+  "description": "compiles tailwind like shorthand syntax into css at runtime",
+  "filename": "twind.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/tw-in-js/twind"
+  },
+  "keywords": [
+    "twind",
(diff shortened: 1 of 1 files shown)


packages/t/twind.json
```diff
@@ -0,0 +1,31 @@
...
+  "autoupdate": {
+    "source": "npm",
+    "target": "twind",
+    "fileMap": [
+      {
+        "basePath": "",
+        "files": [
+          "**/package.json",
+          "**/*.js",
+          "**/*.js.map"
```
MattIPv4, line 26:
```suggestion
          "**/*.js?(.map)"
```
We can reduce this down to a single glob for simplicity

packages/t/twind.json
```diff
@@ -0,0 +1,31 @@
...
+        ]
+      }
+    ]
+  }
+}
```
MattIPv4, line 31:
```suggestion
}

```
Newline at end-of-file :)

MattIPv4 (changes requested):

PR #674, opened 2021-04-02
Fix fingerprintjs2 deprecation notice

Finesse:


Finesse:
Can I trigger the description update on https://cdnjs.com/libraries/fingerprintjs2 without publishing a new version?

MattIPv4:
:wave: It is not possible to update the description until a new version is published, as the change will only be reflected into KV once a new version is detected.

> set up the configuration to take files only from a specific git branch

This is not possible, cdnjs looks for all tags in a GitHub repository or all versions in an NPM package

> redirect from https://cdnjs.com/libraries/fingerprintjs2 to https://cdnjs.com/libraries/@fingerprintjs/fingerprintjs

Unfortauntely, this is also not possible. We will always serve a page on the website for every library that is available via the CDN.

> remove https://cdnjs.com/libraries/fingerprintjs2 (leaving the distributive files for backward compatibility)

See above, every library will have a page.

Finesse:
@maple3142 So could you please just merge the PR? You did it [last time](https://github.com/cdnjs/packages/pull/633).

Will the bot update the page if I delete the latest tag from the repository and add it again?

MattIPv4:
I'm happy to merge a significant change like that to mark a library as deprecated, but to be frank, I'm less inclined to merge a PR that makes a superfluous wording change when its failing CI, and with seemingly little chance of it ever actually making it to KV to be served by the API/website anyway.

> Will the bot update the page if I delete the latest tag from the repository and add it again?

It would not as the bot would have already seen that version name, it would already be present on cdnjs. It would need to be a net-new version, which seems very silly just to trigger a wording change on a deprecated package?

Finesse:
> I'm less inclined to merge a PR that makes a superfluous wording change when its failing CI

I'm sorry to ask for such a daft change, but this particular change is very important:

```diff
- @fingeprintjs/fingerprintjs
+ @fingerprintjs/fingerprintjs
```

Because it's a library identifier, therefore 1 missing letter can break the whole deprecation message.

> seemingly little chance of it ever actually making it to KV to be served by the API/website anyway

What is KV? Fingerprintjs2 is the 88th most popular library on CDNJS according to https://github.com/cdnjs/cf-stats/blob/master/2021/cdnjs_February_2021.md.

MattIPv4:
KV is the key value store where we store all the library metadata for cdnjs -- the metadata stored there is only updated when a new version is published. It looks like you published version 2.1.5 a few hours ago which successfully triggered an update of the KV.

If you're in a position to be able to publish a new version again like that, we can merge this and then do that version release to trigger an update of the KV?

Finesse:
Sure, I realize that the only option I have is publishing a new version. It's ok for me.

MattIPv4 (approved):

Finesse:
@MattIPv4 Thank you 🙏

PR #686, opened 2021-04-08
add human w/ npm auto-update

vladmandic:
new npm package i've released few months ago, just hit v1 recently and already at close to 20k downloads from npmjs  
npmjs


MattIPv4 (approved):
Lgtm -- CI failure seems to be an issue on our end, manually verified this PR

PR #699, opened 2021-04-19
Add juliamono w/ git auto-update

cormullion:
Add juliamono font

### DIFF
--- a/packages/j/juliamono.json
+++ b/packages/j/juliamono.json
@@ -1,7 +1,6 @@
 {
   "name": "juliamono-font",
   "description": "A monospaced font with reasonable Unicode support",
-  "filename": "juliamono.css",
   "keywords": [
     "font",
     "typeface",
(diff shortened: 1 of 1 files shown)


packages/j/juliamono.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "juliamono-font",
+  "description": "A monospaced font with reasonable Unicode support",
+  "filename": "juliamono.css",
```
MattIPv4, line 4:
```suggestion
```
It looks like you're only distributing the font files, so you can probably just omit the default filename?

MattIPv4 (changes requested):

packages/j/juliamono.json
cormullion: Thanks - I wasn't quite sure about this... :)

MattIPv4 (approved):
lgtm, ty!

cormullion:
Thanks Matt!

Issue #555, opened 2020-12-22
prevent the minification (".min.js") files being created?


getify:
My npm packages come with pre-minified files in the `dist/` directory. However, I notice that this cdn automatically builds and adds new files into the directory, with `.min.js` extensions.

Some reasons this is not desired:

* it's unnecessary work
* my `dist/` files all have copyright notices included in them, and your minification is removing those comments -- these copyright notices also include the version of the file, and so with it being removed, I can't verify what version of the code someone might be using.
* it has the potential to cause confusion if someone is using a file from my package that I didn't create and didn't test/verify

I would like to configure my packages to prevent the auto-minification step you're performing. How can I do so?

MattIPv4:
@tc80 could you look at adding this to the bot and package scheme?

tc80:
this is interesting! At the moment, we are *supposedly* ignoring `.js` files that already come with minified forms (right [here](https://github.com/cdnjs/tools/blob/master/compress/js.go#L24-L33)), so this might just be a bug I created...

I'm assuming one of your packages having problems is [CAF](https://github.com/cdnjs/packages/blob/master/packages/c/caf.json)?

@xtuc can you grep the logs for `compressed file already exists`? I think the bug is that `os.Stat` won't return an error if the file exists, so the `!os.IsNotExist(err)` statement is never true. (I'll confirm this in a few minutes)

tc80:
yep. It is just a [bug.](https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go) Whoopsie.

tc80:
@getify this should be fixed soon, please re-raise this issue if you notice auto-minification in your future releases 👍

getify:
I may be misunderstanding the code, but it looks to me like you're checking to see if a file with `.min.js` extension exists... but that's not the way my `dist/` files are organized. I don't distribute any non-minified files, so my files (like `caf.js`) are just always going to have already been minified.

What I'm seeing is *not* that you're overwriting my files, but that you're unnecessarily *adding* `.min.js` files into my `dist/` folder that we're there, and (as far as I'm concerned) shouldn't be there.

tc80:
Ah, gotcha. I just made an issue to add a JSON field to the package config soon for this #556 . I'll tag you when it's done 😄

tc80:
might be a couple weeks because of holidays, but should get fixed 👍 . thanks

KoalaBear84:
[...]

PR #703, opened 2021-04-21
Disabling JS minification for my various packages

getify:
RE: https://github.com/cdnjs/packages/issues/555


getify:
I'm not sure what this check failure is about, as I didn't change anything about the config for CAF other than the minification setting, so this wasn't a problem previously. Please advise if there's something I need to fix about this?

MattIPv4 (approved):
Lgtm

PR #705, opened 2021-04-22
Add Iosevka w/ git auto-update

aymanbagabas:
Use the mirror https://github.com/aymanbagabas/iosevka-fonts instead of https://github.com/be5invis/Iosevka to distribute the fonts.

### DIFF
--- a/packages/i/Iosevka.json
+++ b/packages/i/Iosevka.json
@@ -8,14 +8,20 @@
     "url": "git+https://github.com/be5invis/Iosevka.git"
   },
   "license": "OFL-1.1",
-  "keywords": ["font"],
+  "keywords": [
+    "font"
+  ],
   "autoupdate": {
     "source": "git",
     "target": "git://github.com/aymanbagabas/iosevka-fonts.git",
     "fileMap": [
       {
         "basePath": "dist",
-        "files": ["*/*.css", "*/*/*.(ttf|woff2)"]
+        "files": [
+          "*/*.css",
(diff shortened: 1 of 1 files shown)


packages/i/Iosevka.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "license": "OFL-1.1",
+  "keywords": ["font"],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/aymanbagabas/iosevka-fonts.git",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": ["*/*.css", "*/*/.(ttf|woff2)"]
```
alexeyten, line 18: Shouldn't it be `*/*/*.(ttf|woff2)`?

MattIPv4 (changes requested):
I am very hesitant to accept a PR where the target for auto-updating is known to not be the true source -- there is no history to demonstrate that this source can be trusted, you could easily publish files that are not in the true source.

aymanbagabas:
> I am very hesitant to accept a PR where the target for auto-updating is known to not be the true source -- there is no history to demonstrate that this source can be trusted, you could easily publish files that are not in the true source.

I totally agree. This is related to https://github.com/be5invis/Iosevka/issues/942. The reason for this is that the author refused to maintain the files in the main repository.

MattIPv4 (approved):
Lgtm, seems the original maintainer is okay with this.

PR #706, opened 2021-04-22
Add element-plus w/ npm auto-update

iChenLei:
### New package cdn support
From staticfile's user request https://github.com/staticfile/static/issues/602
`cdnjs` is staticfile's upstream, we can only create new `PR` to `cdnjs` to support `element-plus` cdn feature. 

The official cdn usage.

<del>So i only write two explicit file name</del>
```diff
+         "theme-chalk/index.css",
+         "index.full.js"
```
<del>Will we change the build path  in the future? Is explicit match pattern enough ? </del>
/cc @JeremyWuuuuu @kooriookami @justwiner @zazzaz (active  `element-plus` maintainer)

### Code Review
/cc @MattIPv4  (cdnjs maintainer)

### Appended
<del>My opinion is current config is enough, the `index.full.js` is `UMD` mode, we can update the `element-plus.json` in the future if the build artifacts changed. The `cdnjs` page will show the user needed file cleanly.</del>

### DIFF
--- a/packages/e/element-plus.json
+++ b/packages/e/element-plus.json
@@ -23,7 +23,8 @@
       {
         "basePath": "lib",
         "files": [
-          "theme-chalk/index.css",
+          "theme-chalk/*.css",
+          "theme-chalk/fonts/*.@(ttf|woff)",
           "index.full.js"
         ]
       }
--- a/packages/e/element-plus.json
+++ b/packages/e/element-plus.json
@@ -1,0 +1,46 @@
+{
+  "name": "element-plus",
+  "filename": "index.full.js",
+  "description": "A Vue.js 3.0 UI Library made by Element team",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/element-plus/element-plus.git"
+  },
+  "license": "MIT",
+  "homepage": "https://element-plus.org",
+  "keywords": [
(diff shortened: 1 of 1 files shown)


packages/e/element-plus.json
```diff
@@ -0,0 +1,46 @@
...
+    "vue-composition-api"
+  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "element-plus",
+    "fileMap": [
+      {
+        "basePath": "lib",
+        "files": [
+          "theme-chalk/index.css",
```
MattIPv4, line 26:
```suggestion
          "theme-chalk/*.css",
          "theme-chalk/fonts/*.@(ttf|woff)",
```
We should support all the css files that the library provides, for if users only want certain styles.
You also need to include the fonts.

packages/e/element-plus.json
```diff
@@ -0,0 +1,46 @@
...
+          "index.full.js"
```
MattIPv4, line 27:
```suggestion
          "index.*.js"
```
We should also support the esm option

MattIPv4 (changes requested):

MattIPv4 (approved):
Lgtm, ty

PR #721, opened 2021-05-06
Update juliamono w/ git auto-update

cormullion:
Adding CSS files.

### DIFF
--- a/packages/j/juliamono.json
+++ b/packages/j/juliamono.json
@@ -14,7 +14,7 @@
   },
   "license": "OFL-1.1",
   "homepage": "https://github.com/cormullion/juliamono",
-  "filename": "webfonts/juliamono-regular.css",
+  "filename": "juliamono-regular.css",
   "autoupdate": {
     "source": "git",
     "target": "git://github.com/cormullion/juliamono.git",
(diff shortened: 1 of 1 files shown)


packages/j/juliamono.json
```diff
@@ -14,14 +14,16 @@
   },
   "license": "OFL-1.1",
   "homepage": "https://github.com/cormullion/juliamono",
+  "filename": "webfonts/juliamono-regular.css",
```
mortenpi, line 17: Should it maybe be this, because `basePath = "webfonts"`?

```suggestion
  "filename": "juliamono-regular.css",
```

cormullion: That might be a better guess than mine! :)

cormullion:
Thanks!

cormullion:
@MattIPv4 Do I have to do something else to get this package to auto-update?

MattIPv4:
👋 We have ongoing maintenance with the auto-update functionality of cdnjs currently -- this change was made since the maintenance started, and so our auto-update logic hasn't been running to detect it. You can follow along on our status page for when the maintenance will conclude: https://status.cdnjs.com/

cormullion:
Ok, thanks!

PR #723, opened 2021-05-06
Add google-libphonenumber w/ npm auto-update

ruimarinho:
Adding google-libphonenumber using npm auto-update. Pre-compiled, ready-to-use wrapper for libphonenumber from Google.

### DIFF
--- a/packages/g/google-libphonenumber.json
+++ b/packages/g/google-libphonenumber.json
@@ -1,0 +1,41 @@
+{
+  "description": "The up-to-date and reliable Google's libphonenumber package for node.js for parsing, formatting, and validating international phone numbers.",
+  "filename": "libphonenumber.js",
+  "homepage": "https://github.com/ruimarinho/google-libphonenumber",
+  "keywords": [
+    "e164",
+    "format",
+    "formatting",
+    "international",
+    "libphonenumber",
+    "number",
+    "phone",
+    "phonenumber",
+    "rfc3966",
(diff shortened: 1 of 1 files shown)


packages/g/google-libphonenumber.json
```diff
@@ -0,0 +1,41 @@
+{
+  "description": "The up-to-date and reliable Google's libphonenumber package for node.js for parsing, formatting, and validating international phone numbers.",
+  "filename": "dist/libphonenumber.js",
```
MattIPv4, line 3:
```suggestion
  "filename": "libphonenumber.js",
```
As the base dir is `dist`, we'll pull the files out of that directory into the root, so it isn't needed here.

packages/g/google-libphonenumber.json
```diff
@@ -0,0 +1,41 @@
...
+    "url": "https://github.com/ruimarinho/google-libphonenumber.git"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "google-libphonenumber",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 30:
```suggestion
          "*.js"
```
We try to avoid globstars when possible, especially open ones -- it looks like we just need to match the JS files here so this should be sufficient.

MattIPv4 (changes requested):

PR #747, opened 2021-06-06
Update highlight.js w/ git auto-update

joshgoebel:
Also, the CDN doesn't seem to be picking up our three latest versions/tags.

- 11.0.0
- 11.0.1
- 10.7.3

Could anyone help?  Our build process hasn't change or anything, we push releases to our CDN repo just like we always have...

### DIFF
--- a/packages/h/highlight.js.json
+++ b/packages/h/highlight.js.json
@@ -23,7 +23,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/isagalaev/highlight.js"
+    "url": "https://github.com/highlightjs/highlight.js"
   },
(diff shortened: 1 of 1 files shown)


joshgoebel:
Is it possible to use NPM as a source rather than git?

joshgoebel:
Related: #748

MattIPv4:
👋 We have ongoing maintenance with the auto-update functionality of cdnjs currently -- these versions were released since the maintenance started, and so our auto-update logic hasn't been running to detect them. You can follow along on our status page for when the maintenance will conclude: https://status.cdnjs.com/

cc @xtuc

MattIPv4 (approved):

PR #748, opened 2021-06-06
Update highlight.js w/ npm auto-update

joshgoebel:
I don't know how to test this, but at a glance it looks correct.

### DIFF
--- a/packages/h/highlight.js.json
+++ b/packages/h/highlight.js.json
@@ -17,7 +17,7 @@
           "highlight.min.js",
           "languages/*.min.js",
           "styles/*",
-          "styles/**/*"
+          "styles/**/*.css"
         ]
       }
     ]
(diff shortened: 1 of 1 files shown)


packages/h/highlight.js.json
```diff
@@ -1,22 +1,23 @@
...
     "fileMap": [
       {
-        "basePath": "build",
+        "basePath": "",
         "files": [
           "highlight.min.js",
           "languages/*.min.js",
-          "styles/*"
+          "styles/*",
+          "styles/**/*"
```
MattIPv4, line 20:
```suggestion
          "styles/**/*.css"
```
Unrestricted globstars are dangerous and we try to avoid them by ensuring we have extensions etc. to limit them.

MattIPv4 (changes requested):

MattIPv4 (approved):
Lgtm

joshgoebel:
Will the missing versions magically catch-up I assume when the auto-updater comes back online?

MattIPv4:
Yes, I am under the impression that the team will work to backfill missing versions once the new system is online.

PR #756, opened 2021-06-16
Update ag-grid w/ git auto-update

seanlandsman:


packages/a/ag-grid.json
```diff
@@ -4,36 +4,44 @@
...
   ],
   "license": "MIT",
   "homepage": "http://www.ag-grid.com/",
   "autoupdate": {
     "source": "git",
-    "target": "git://github.com/ceolter/ag-grid.git",
+    "target": "git://github.com/ag-grid/ag-grid.git",
     "fileMap": [
       {
         "basePath": "dist",
```
MattIPv4, line 34: I'm guessing this might need to become `community-modules/all-modules/dist`, based on https://github.com/ag-grid/ag-grid/tree/v25.3.0/community-modules/all-modules/dist

seanlandsman: Thanks @MattIPv4 - are you saying the "basePath" needs to change? Can I ask where cdnjs looks for these files? I ask as we use release branches (and tag) them, so we don't use master/main for this, or the default branch (latest). 

Apologies if this is documented somewhere - I did look but was unable to find anything on how to configure the json file

thanks!

MattIPv4: 👋 Git-based auto-update is done from git tags -- I've linked to your lastest tag above, and what I believe the new `basePath` needs to be to get the new files that should be distributed, unless those aren't what should be distributed?

seanlandsman: Thanks, I've updated it with what the path should be (you were close :-) ) - thanks for your help

MattIPv4 (approved):
Lgtm, tyvm!

PR #761, opened 2021-06-20
Update mustache.js w/ npm auto-update

phillipj:
Previously the mustache git repository had its `mustache.js` source code written in UMD.

Recently a build step was introduced, where `mustache.js` in the git repository is now written in ESM syntax, but converted into UMD in the build process before being published to npm.

The ESM syntax alternative is still available as `mustache.mjs` in the npm package.

Refs https://github.com/janl/mustache.js/commit/cc979e0419e7a1aae6c14c1a59763324855a54f3#commitcomment-52315118

/cc @hammond13

### DIFF
--- a/packages/m/mustache.js.json
+++ b/packages/m/mustache.js.json
@@ -10,14 +10,15 @@
     "ejs"
   ],
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/janl/mustache.js.git",
+    "source": "npm",
+    "target": "mustache",
     "fileMap": [
       {
         "basePath": "",
         "files": [
           "mustache.js",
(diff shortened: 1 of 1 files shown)


MattIPv4 (approved):
Lgtm, tyvm

phillipj:
Thanks @MattIPv4!

Will these changes affect the already published v4.2.0 version of mustache on cdnjs.com at some point? Or will it be as is until a new version of mustache is published to npm?

MattIPv4:
Once a version is published on the CDN it will never be changed, so these changes will only affect future versions that are published.

phillipj:
Got it, thanks for the lightning quick response 👍🏻

PR #775, opened 2021-07-14
Add epijs w/ git auto-update

Quantalabs:
Adds the EpiJS package to cdnjs

### DIFF
--- a/packages/e/epijs.json
+++ b/packages/e/epijs.json
@@ -1,5 +1,5 @@
 {
-  "name": "EpiJS",
+  "name": "epijs",
   "description": "A toolset for analysing and creating epidemiological models",
   "filename": "index.min.js",
   "repository": {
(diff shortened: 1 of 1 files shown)


MattIPv4 (approved):
Lgtm

Quantalabs:
Is it just my end, or does cdnjs only show the `pre.min.js` file, when it should be showing the `index.min.js` file?

MattIPv4:
It looks like the 1.3.x versions haven't been loaded yet -- the latest version on the site currently is [1.2.1](https://cdnjs.com/libraries/epijs/1.2.1), which has the correct files from that tag in your repo: https://github.com/Quantalabs/EpiJS/tree/v1.2.1/web

Quantalabs:
Oh, I see, I still don't think that the loading of the other files are necessary, and fixed it in #780, and I also fixed a typo.

PR #780, opened 2021-07-17
Update epijs w/ git auto-update

Quantalabs:
Correct filename in the EpiJS package file, to stop loading of the wrong file on the EpiJS CDNJS page.


MattIPv4 (approved):
Lgtm

PR #850, opened 2021-09-13
Update noUiSlider w/ npm auto-update

pableu:
See their [CHANGELOG.MD](https://github.com/leongersen/noUiSlider/blob/master/CHANGELOG.MD) for 15.0.0: 

> The distributed files have moved from distribute to dist in the NPM package, and are no longer in the repository; You may need to change the path to the noUiSlider CSS file if you are importing it using a package manager;

I'm not 100% sure if I got the syntax right.

And `"basePath": "dist"` isn't backwards compatible to versions 14.x and lower, but I guess that won't change the versions already in cdnjs.

### DIFF
--- a/packages/n/noUiSlider.json
+++ b/packages/n/noUiSlider.json
@@ -15,11 +15,11 @@
     "slide"
   ],
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/leongersen/noUiSlider.git",
+    "source": "npm",
+    "target": "nouislider",
     "fileMap": [
       {
-        "basePath": "distribute",
+        "basePath": "dist",
         "files": [
           "**/*"
         ]


MattIPv4 (approved):
LGTM, tyvm!

pableu:
So how long until new versions are visible in https://cdnjs.com/libraries/noUiSlider?

pableu:
@MattIPv4 I'm sorry, but I think this didn't work out as expected. 

The version in https://cdnjs.com/libraries/noUiSlider still 14.7.0.  But the search result says  15.3.0.
image

And in the cdnjs/cdnjs repo, we also only have 14.7.0: https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/noUiSlider

Can you take a look again, maybe? :-)

MattIPv4:
Looks like this got caught by a bug a while back where they were published as empty versions: https://github.com/cdnjs/logs/tree/prod/packages/n/noUiSlider -- will look to see if I can force them to try again

pableu:
Yup, it seems you fixed it: https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/noUiSlider/15.4.0 Thanks :-)

MattIPv4:
Most welcome :)

PR #864, opened 2021-10-06
Add Animate.js w/ npm auto-update

koffiisen:


MattIPv4:
👋 I'm a bit concerned that this has a naming collision with https://www.npmjs.com/package/animate.js which is an established package?

MattIPv4:
Given your package currently only has 6 stars on GitHub, and is not published on GitHub, it does not fall within our [popularity requirement](https://github.com/cdnjs/packages/blob/master/CONTRIBUTING.md#policy-rules-and-guidelines). Further, I'm also concerned about future issues this could cause as it is a naming collision with an established NPM package. As such, closing.

PR #870, opened 2021-10-15
Add jose w/ git auto-update

### LINKED ISSUE #869: Add: jose
### Library name

jose

### Library license

https://github.com/panva/jose/blob/main/LICENSE.md

### Library source

https://github.com/panva/jose/tree/v4.0.2

---

Notes:

- only tagged commits include the `dist` folder
- the browser entry point is only `./dist/browser/index.js`
- types (should they be applicable) are in `./dist/types/index.d.ts`

panva:
Adding jose using git auto-update from git://github.com/panva/jose.git

resolves #869

### DIFF
--- a/packages/j/jose.json
+++ b/packages/j/jose.json
@@ -1,5 +1,6 @@
 {
   "name": "jose",
+  "filename": "index.js",
   "description": "'JSON Web Almost Everything' - JWA, JWS, JWE, JWT, JWK, JWKS with no dependencies using runtime's native crypto",
   "keywords": [
     "compact",


panva:
> Looking at your browser dist directory, it seems that this is ESM code? Is this project something that will only work in browsers that support modules?

correct

> (Fwiw, I was expecting to find a single, bundled file with cjs/umd support)

I can add a single ESM bundle

panva:
> Is `index.js` the recommended default file for folks to use?

yes

panva:
@MattIPv4 thank you for walking me through this.

May I inquire about the frequency of auto-updates, and if I need to publish a new release now to have the package published?

MattIPv4:
You shouldn't need to publish a new version, we'll detect the most recent few you already have. The auto-update runs every 2 hours or so iirc, as it takes about that long to check all the packages we have.

PR #893, opened 2021-11-12
Update vue-tour w/ npm auto-update

smitpatelx:
- Updated the package description from the official site.
- Added css files in CDN

### Issue:
- Your site still shows the old github url (Gihub link on https://cdnjs.com/libraries/vue-tour) instead of the official one you see above.


MattIPv4 (approved):
lgtm, ty

PR #931, opened 2021-12-21
Add amis w/ npm auto-update

nwind:


MattIPv4 (changes requested):
👋 `amis/lib/index.js` relies on Node.js `require` calls to load the library -- this is not compatible with using a JS library on the web, which is the primary purpose that cdnjs serves.

Is there a version of this that is pre-bundled and would work on the web that should be used instead?

MattIPv4:
No progress, closing.

PR #980, opened 2022-01-22
Update js-cookie w/ npm auto-update

### LINKED ISSUE #979: Update: js-cookie is not updating. Latest version is 3.0.1
### Library name

js-cookie

### Library JSON file

https://github.com/cdnjs/packages/blob/master/packages/j/js-cookie.json

### Library source

https://www.npmjs.com/package/js-cookie

### Files to include

- js.cookie.min.js
- js.cookie.js
- js.cookie.min.mjs
- js.cookie.mjs

One difference I found is that it changed the default branch from `master` to `main`

MattIPv4:
Resolves #979 (hopefully, though I see nothing wrong with the config currently)

### DIFF
--- a/packages/j/js-cookie.json
+++ b/packages/j/js-cookie.json
@@ -19,7 +19,7 @@
       {
         "basePath": "dist",
         "files": [
-          "*.js"
+          "*.@(js|mjs)"
         ]
       }
     ]


MattIPv4 (commented):
lgtm

luislobo:
Do you happen to know how much time does it take to be publicly available? 

It is still showing 
image

Latest version is 3.0.1 on npmjs

PR #1040, opened 2022-03-02
Include textarea for files in library request

MattIPv4:
As is already present in the update request template, include a textarea in the library request template for users to list out what files should be included when a library is requested to be added to cdnjs.


MattIPv4 (commented):
lgtm

PR #1043, opened 2022-03-04
Add stanford-slate-enhancements w/ git auto-update

redndahead:


MattIPv4:
👀 It looks like our git support is only set up to support GitHub at present (it appears to be fetching tags via the GitHub API, rather than via pure git).

So, unfortunately, I'm not going to be able to accept this request at present. I recommend opening an issue in the tools repo to request support for git support out of GitHub if you wish :) Sorry!

(for refL https://github.com/cdnjs/tools/blob/e87c098c9dc2dce6b3f49c77adf4f29b60eb5366/cmd/checker/main.go#L135 + https://github.com/cdnjs/tools/blob/e87c098c9dc2dce6b3f49c77adf4f29b60eb5366/git/github.go#L102-L106)

MattIPv4:
Alternatively (and what I'd recommend), you can look at distributing your package through NPM which'd allow us to make a proper judgement call on package popularity, and is our preferred method for fetching library versions to host :) It'd also mean that folks can use your package throughout the wider JS ecosystem, not just via us!

PR #1112, opened 2022-04-21
Update react-inputs-validation w/ npm auto-update

edwardfxiao:
Update the ```homepage```, ```repository URL```, and ```description``` of``` react-inputs-validation.json``` since the GitHub username has changed.

### DIFF
--- a/packages/r/react-inputs-validation.json
+++ b/packages/r/react-inputs-validation.json
@@ -1,5 +1,5 @@
 {
-  "description": "a react inputs validation component",
+  "description": "A react component for form inputs validation.",
   "filename": "react-inputs-validation.min.js",
   "homepage": "https://edwardfxiao.github.io/react-inputs-validation/",
   "keywords": [
--- a/packages/r/react-inputs-validation.json
+++ b/packages/r/react-inputs-validation.json
@@ -1,7 +1,7 @@
 {
   "description": "a react inputs validation component",
   "filename": "react-inputs-validation.min.js",
-  "homepage": "https://edwardfhsiao.github.io/react-inputs-validation/",
+  "homepage": "https://edwardfxiao.github.io/react-inputs-validation/",
   "keywords": [
     "react",
     "input",
@@ -19,7 +19,7 @@
   "name": "react-inputs-validation",
   "repository": {
     "type": "git",
-    "url": "https://github.com/edwardfhsiao/react-inputs-validation"
+    "url": "https://github.com/edwardfxiao/react-inputs-validation"
   },
   "autoupdate": {
     "source": "npm",


MattIPv4 (approved):
LGTM

edwardfxiao:
@MattIPv4 Hi, I don't see these two change
image
and the https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/react-inputs-validation/package.json has not been updated either.

Is there anything else I should do to get them updated?

MattIPv4:
The data won't update until a new version of the library is published :)

edwardfxiao:
@MattIPv4 Thank you for the clarification

edwardfxiao:
@MattIPv4
I just updated a new version of 4.9.2 and https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/react-inputs-validation has been updated but [package.json](https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/react-inputs-validation/package.json) still has not changed, any suggestion?

MattIPv4:
I don't think we update those top-level JSON files anymore, but you can see that the updated JSON is present for the new version: https://github.com/cdnjs/logs/commit/b061c3c090eb6f9213a76d463599497cdf06d5c9

edwardfxiao:
> I don't think we update those top-level JSON files anymore

The wrong top-level JSON file takes my ```homepage``` to a ```404``` page. It would be nice if there is a way to correct it,
I wonder if the team will have any plan for this?

edwardfxiao:
To me, [cdnjs.com](https://cdnjs.com/) is still a very serious official site for libraries, and there are many other websites all over the world that are mirroring from it. 

So I personally would love it keep the information accurate.

MattIPv4:
Yes, I agree -- cdnjs.com is driven by data in our KV store, not by the cdnjs/cdnjs repository.

edwardfxiao:
@MattIPv4 Sorry I was confused about the relationships of ```cdn/cdn``` ```cdn website```

I understand what you saying now.

- the top-level [package.js](https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/react-inputs-validation/package.json) on ```cdn/cdn``` repository supposedly doesn't have an effect on [cdnjs.com](https://cdnjs.com/) anymore because it is more like a backup now
- the data in KV store takes over the responsibility to the [cdnjs.com](https://cdnjs.com/), and the log does show the package.json file has been updated

Interestingly that the ```description``` on the response of the search box has been updated whereas on the detail page doesn't.

Do you think I could open an issue to track this?





image

MattIPv4:
Sure -- please open it in cdnjs/api-server (it looks like https://api.cdnjs.com/libraries/react-inputs-validation doesn't have the new info)

PR #1123, opened 2022-04-26
Update waline w/ npm auto-update

leirock:
v2 also provides files in other formats rather than only .js files.

### DIFF
--- a/packages/w/waline.json
+++ b/packages/w/waline.json
@@ -24,5 +24,6 @@
         ]
       }
     ]
-  }
+  },
+  "filename": "waline.js"
 }
--- a/packages/w/waline.json
+++ b/packages/w/waline.json
@@ -1,30 +1,30 @@
 {
-  "name": "waline",
-  "description": "client for waline comment system",
-  "keywords": [
-    "valine",
-    "waline",
-    "comment",
-    "blog"
-  ],
-  "homepage": "https://waline.js.org/",
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/walinejs/waline.git"
-  },
-  "license": "MIT",
-  "autoupdate": {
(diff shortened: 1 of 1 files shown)


packages/w/waline.json
```diff
@@ -20,10 +20,9 @@
       {
         "basePath": "dist",
         "files": [
-          "*.js"
+          "**/*"
```
MattIPv4, line 23: 👋 We try to avoid using globstars as they can very easily lead to abuse of the CDN -- please try to specify what files we're expecting to match based on their extensions and directories.

leirock: ok, I will add another commit to specify files later.

leirock:
I just listed three files in the new commit.

packages/w/waline.json
```diff
@@ -20,10 +20,11 @@
       {
         "basePath": "dist",
         "files": [
-          "*.js"
+          "waline.css",
+          "waline.js",
+          "pageview.js"
```
MattIPv4, line 25:
```suggestion
          "*.@(js|css)"
```
It looks like there are quite a few other JS files beyond just `waline.js` and `pageview.js` that other developers may want to us, so let's select every JS/CSS file

packages/w/waline.json
```diff
@@ -20,10 +20,11 @@
...
         ]
       }
     ]
-  },
-  "filename": "Waline.min.js"
+  }
```
MattIPv4, line 29:
```suggestion
  },
    "filename": "waline.js"
```
We can still keep the filename as the recommended file for folks to use

MattIPv4 (changes requested):

MattIPv4 (approved):
LGTM

PR #1137, opened 2022-05-04
Update europa w/ npm auto-update

neocotic:
Europa's configuration has changed post-v5 due to a move in repositories as well as a change in the structure of its distribution files.

### DIFF
--- a/packages/e/europa.json
+++ b/packages/e/europa.json
@@ -10,15 +10,16 @@
   ],
   "repository": {
     "type": "git",
-    "url": "https://github.com/NotNinja/europa.git"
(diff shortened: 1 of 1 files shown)


MattIPv4 (approved):
LGTM, ty

neocotic:
Thanks @MattIPv4. Do you know when I'm likely to see these changes here: https://cdnjs.com/libraries/europa

MattIPv4:
The auto-update changes will happen when the next update gets published -- as for the top-level info changes, I think we might have a bug that prevents those updating, or they may also update with the next release of the library.

PR #1143, opened 2022-05-07
Update xlsx w/ git auto-update

jameshilliard:
This package no longer receives updates on npm

### DIFF
--- a/packages/x/xlsx.json
+++ b/packages/x/xlsx.json
@@ -13,11 +13,11 @@
   ],
   "repository": {
     "type": "git",
-    "url": "https://github.com/SheetJS/js-xlsx.git"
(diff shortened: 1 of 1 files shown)


MattIPv4:
@xtuc Any ideas on solving for `2022/05/08 13:14:01 failed to process version: failed to run sandbox: could not create container: Error response from daemon: Invalid container name (xlsx_0.18.4+deno), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed`?

MattIPv4:
While the CI is now passing here (woo!), it looks like the repository this PR is changing the package to point at is also abandoned, and all the tagged versions that exist on it also exist on NPM. As such, closing.

PR #1173, opened 2022-06-01
Update jodit w/ npm auto-update

xdan:
I want to separate some of the plugins from the general build so as not to bloat the main file
Therefore, it is necessary that internal folders in build be laid out in cdn

https://github.com/xdan/jodit/tree/master/build plugins folder

### DIFF
--- a/packages/j/jodit.json
+++ b/packages/j/jodit.json
@@ -21,7 +21,8 @@
       {
         "basePath": "build",
         "files": [
-          "jodit*.+(css|js)"
+          "jodit*.+(css|js)",
+          "**/*"
         ]
       }
     ]
--- a/packages/j/jodit.json
+++ b/packages/j/jodit.json
@@ -22,7 +22,7 @@
         "basePath": "build",
         "files": [
           "jodit*.+(css|js)",
-          "plugins/**/*.(css|js)"
+          "plugins/**/*.@(css|js)"
         ]
       }
     ]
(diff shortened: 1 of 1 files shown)


packages/j/jodit.json
```diff
@@ -21,7 +21,8 @@
       {
         "basePath": "build",
         "files": [
-          "jodit*.+(css|js)"
+          "jodit*.+(css|js)",
+          "**/*"
```
MattIPv4, line 25: 👋 We generally try to avoid using open globstars in file maps as it can lead to files being matched unintentionally that shouldn't be on the CDN, causing performance impact to our pipelines and making the library responses on the site/API confusing for users.

Could this be scoped to just what you intend to match? Looking at your repo, I suspect this might just need to be `plugins/**/*.@(css|js)`?

MattIPv4 (changes requested):

packages/j/jodit.json
xdan: Hi. Thank you. Corrected. Just did not understand about @, is this symbol definitely needed?

MattIPv4: `@(css|js)` means `css or js` in glob syntax :)

MattIPv4: You can play around with different glob patterns at https://www.digitalocean.com/community/tools/glob if you want

xdan: Thank you. Returned the @ sign)

MattIPv4 (approved):
LGTM

PR #1205, opened 2022-06-15
The project `flag-icon-css` have been renamed to `flag-icons`

lipis:
I'm the author: https://github.com/lipis/flag-icons


MattIPv4 (changes requested):
:wave: Hey! Thanks for creating the PR. Unfortunately it's not possible to directly rename a package on cdnjs (renaming the package config file would essentially result in all the old versions disappearing).

Instead, my recommendation would be to keep the old file as-is (potentially updating the autoupdate target, and/or adding a note to the description), and then adding the project under the new name as a new package config.

lipis:
OK.. I'll create a new PR

PR #1226, opened 2022-07-06
Add wangeditor5 w/ npm auto-update

wangfupeng1988:


packages/w/wangeditor5.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "wangEditor5",
```
MattIPv4, line 2:
```suggestion
  "name": "wangeditor5",
```
The `name` property should match the file name.

packages/w/wangeditor5.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "license": "MIT",
+  "autoupdate": {
+    "source": "npm",
+    "target": "@wangeditor/editor",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 23:
```suggestion
          "*.js?(.map)",
          "css/*.css"
```
We try to avoid using open globstars as they can result in matching files that we didn't intend to include.

MattIPv4 (changes requested):

MattIPv4 (approved):
LGTM, ty

PR #1236, opened 2022-07-17
Update discuss w/ npm auto-update

Lete114:


packages/d/discuss.json
```diff
@@ -8,28 +8,35 @@
...
   "autoupdate": {
     "source": "npm",
     "target": "discuss",
     "fileMap": [
       {
-        "basePath": "dist",
+        "basePath": "",
         "files": [
-          "*.js"
+          "dist/*",
```
MattIPv4, line 35:
```suggestion
          "dist/*.js",
```
We try to avoid open patterns as this can result in us matching files we don't want.

packages/d/discuss.json
```diff
@@ -8,28 +8,35 @@
...
+          "assets/*"
```
MattIPv4, line 36:
```suggestion
          "assets/@(emot|img)/*.png",
          "assets/svg/*.svg"
```
I don't see any files in the top-level assets directory, I suspect you need a more specific glob here to match the files you actually want.

packages/d/discuss.json
```diff
@@ -8,28 +8,35 @@
...
         ]
       }
     ]
   },
-  "filename": "Discuss.js"
+  "filename": "discuss.js"
```
MattIPv4, line 41:
```suggestion
  "filename": "dist/discuss.js"
```
As you've updated the fileMap, this file is now in the `dist` subdirectory

MattIPv4 (changes requested):

packages/d/discuss.json
MattIPv4: Ah okay, I would suggest using `dist/*.js?(.map)` then :)

MattIPv4 (approved):
LGTM, ty

PR #1253, opened 2022-07-28
Add webcomponents-template-polyfill w/ npm auto-update

renbaoshuo:


packages/w/webcomponents-template-polyfill.json
```diff
@@ -0,0 +1,46 @@
...
+    "type": "git",
+    "url": "git+https://webcomponents@github.com/webcomponents/polyfills.git"
+  },
+  "filename": "template.min.js",
+  "autoupdate": {
+    "source": "npm",
+    "target": "@webcomponents/template",
+    "fileMap": [
+      {
+        "basePath": ".",
```
MattIPv4, line 33:
```suggestion
        "basePath": "",
```
We can leave this blank for the root :)

packages/w/webcomponents-template-polyfill.json
```diff
@@ -0,0 +1,46 @@
...
+        "files": [
+          "*.@(js|ts|map)"
```
MattIPv4, line 35:
```suggestion
          "*.js?(.map)",
          "externs/*.d.ts"
```
It looks like your typings are in a subdirectory, so that'll need to be specified here

MattIPv4 (changes requested):

MattIPv4 (approved):
LGTM, ty

PR #1263, opened 2022-08-01
Add albe-web-components w/ npm auto-update

saldomik:
Adds albe web components - javascript web components library - to cdnjs

https://github.com/ZanichelliEditore/design-system

https://www.npmjs.com/package/@zanichelli/albe-web-components

### DIFF
--- a/packages/a/albe-web-components.json
+++ b/packages/a/albe-web-components.json
@@ -1,0 +1,27 @@
+{
+    "name": "albe-web-components",
+    "filename": "albe-web-components.min.js",
+    "description": "Zanichelli Editore Design System library",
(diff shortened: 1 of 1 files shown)


saldomik:
Hi :hand:, it seems that the pipeline fails because it cannot find the file albe-web-components.min.js. According to the documentation, it should self-generate. Also, the pipeline seems to have been running for 6 hours non-stop.

MattIPv4:
👀 Could I suggest replacing the globstar in your config with more specific entries for the files you need? The globstar matches over 1000 files which is likely part of the reason the CI took so long, as it had to process each and every one of them.

MattIPv4:
No progress, closing.

PR #1181, opened 2022-06-07
Add jssm w/ npm auto-update

StoneCypher:
Adds `jssm` - javascript state machine - to the CDN

https://www.npmjs.com/package/jssm

https://github.com/StoneCypher/jssm

Thank you for a helpful system

----

Resolves: #1182

PR #1266, opened 2022-08-09
Update vditor w/ npm auto-update

am-abudu:


MattIPv4:
👋 Please could you provide some context on why this change is needed?

MattIPv4:
Ah, interesting. An odd requirement to have, but the change makes sense in that context. Thanks.

MattIPv4:
Looks like the CI is timing out here, similar to #1181 -- this has been escalated to the Cf folks that own the CI flow.

Issue #926, opened 2021-12-14
Remove or unpublish prism@9000.0.1


jimmylewis:
If you search for prism, the default version returned is 9000.0.1.  This is a mysterious package version; nobody knows where it came from, and it was removed from NPM back in 2019 (see https://securityreport.com/where-did-these-mysterious-prismjs-npm-versions-come-from/ for an account of the debacle).

The 9000 version is an unknown asset and should not be served to anyone.

Currently, this version is shown when initially [searching for prism](https://cdnjs.com/libraries?q=prism), or by viewing the library catalog (e.g. https://api.cdnjs.com/libraries?fields=name,version)
```
...
{
    "name": "prism",
    "latest": "https://cdnjs.cloudflare.com/ajax/libs/prism/9000.0.1/prism.min.js",
    "version": "9000.0.1"
},
...
```

Other tools which use this API to show latest versions could be sending users into a bad outcome.

MattIPv4:
cdnjs has a very strict policy against un-publishing existing assets on the CDN, as folks may be using them directly, or be reliant on the data for them existing in the API.

That being said, I agree with working to make this not the default version recommended for the library. @xtuc / @tc80 any thoughts on the best way we can change the default version for this library? iirc we so semver sorting for auto-update, so I imagine this'll need some level of manual intervention to override?

RunDevelopment:
I am one of Prism's maintainers and state my support for removing this version or at least not making it the default version [here](https://github.com/cdnjs/cdnjs/discussions/14145#discussion-4211708). 

This version causes problem because some services to update dependencies use cdnJS's database to determine the latest version of a package. v9000.0.1 not only causes these services to not update Prism, they also recommend "updating" to an old, vulnerable version of Prism.

Ideally, this version should be removed outright. Nobody should be relying on this.

ycw:
> cdnjs has a very strict policy against un-publishing existing assets on the CDN, as folks may be using them directly, or be reliant on the data for them existing in the API.

@MattIPv4 reasonable ... but why the stats (of 0 vulnerability) is linking from v1.28.0 report while v9000.0.1 is autopicked??

MattIPv4:
> but why the stats (of 0 vulnerability) is linking from v1.28.0 report while v9000.0.1 is autopicked??

[...]

PR #1277, opened 2022-08-15
Update react-router w/ npm auto-update

renbaoshuo:


MattIPv4:
cc #1289 #926 -- looks like we've got an existing use case of `ignoreVersions` in a config, a property we don't currently support

renbaoshuo:
The location of `react-router.production.min.js` has changed since `react-router@6.4.0`. I just updated it and hope you will review this PR again.

renbaoshuo:
@MattIPv4

MattIPv4 (approved):
LGTM, ty (and looks like the removal of the `ignoreVersions` is safe now, the `latest` version doesn't seem to be on NPM now)

PR #1309, opened 2022-08-23
Update goodshare.js w/ git auto-update

Asperhsu:


MattIPv4 (changes requested):
👋 I'd love to know a bit more about the context here -- I don't see anything indicating that this new project is the same project as the old one?

PR #1327, opened 2022-09-17
Add miaoyan w/ git auto-update

tw93:


packages/m/miaoyan.json
```diff
@@ -0,0 +1,33 @@
...
+  },
+  "license": "MIT",
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/tw93/MiaoYan.git",
+    "fileMap": [
+      {
+        "basePath": "Resources",
+        "files": [
+          "**/*"
```
MattIPv4, line 22: 👋 We try to avoid open globstars in these configs, as they can lead to matching files that we don't expect to match. Please be more specific with your globs.

packages/m/miaoyan.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "miaoyan",
+  "description": "Lightweight Markdown app to help you write great sentences.",
+  "filename": "index.min.js",
```
MattIPv4, line 4: This filename should match a file in your project, the default file that folks should be importing into a browser to start using your project.

MattIPv4 (changes requested):

packages/m/miaoyan.json
tw93: done

tw93:
I have finished the modification. Please review～

packages/m/miaoyan.json
```diff
@@ -0,0 +1,33 @@
...
+  },
+  "license": "MIT",
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/tw93/MiaoYan.git",
+    "fileMap": [
+      {
+        "basePath": "Resources",
+        "files": [
+          "**/*.+(js|css|ttf)"
```
MattIPv4, line 22: Can we avoid using a globstar here please, and instead specify which directories assets should be pulled from?

packages/m/miaoyan.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "miaoyan",
+  "description": "Lightweight Markdown app to help you write great sentences.",
+  "filename": "package.json",
```
MattIPv4, line 4: You are only including js/css/ttf files -- `filename` should be the name of one of the files you are including, the base file that users should import into their site to use your project

MattIPv4 (changes requested):

MattIPv4:
No progress, closing.

PR #1361, opened 2022-10-06
Update alasql w/ npm auto-update

mathiasrw:
The repo for AlaSQL was moved to the AlaSQL organisation on Github.

### DIFF
--- a/packages/a/alasql.json
+++ b/packages/a/alasql.json
@@ -30,8 +30,8 @@
   ],
   "license": "MIT",
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/AlaSQL/alasql.git",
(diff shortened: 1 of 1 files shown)


MattIPv4 (changes requested):
👋 It looks like the repository (https://github.com/AlaSQL/alasql/tree/v2.1.3) no longer contains the `dist` directory -- should this be switched over to pulling from NPM instead of Git?

mathiasrw:
Ahhh - good point! Ill try to make it point to NPM

mathiasrw:
Ok - I think I got it. Please review and let me know if I should do better...

MattIPv4 (approved):
LGTM, ty

PR #1367, opened 2022-10-11
Add soroban-client w/ npm auto-update

paulbellamy:


packages/s/soroban-client.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "license": "Apache-2.0",
+  "autoupdate": {
+    "source": "npm",
+    "target": "soroban-client",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 24: 👋 We try to avoid using globstars, especially open ones, as they can lead to matching files that we didn't intend to match. Please can you specify the directories and extensions that you want to match here -- looking at the CI run, probably just `*.js` would do.

MattIPv4 (changes requested):

MattIPv4:
No progress, closing.

paulbellamy:
Was away for a few days. I've fixed the globstar thing. Can we re-open this?

PR #1369, opened 2022-10-13
Add carrierjs w/ npm auto-update

tamalchowdhury:
I added a / to `basePath` since I don't know what to do when the file is directly under the package and not under `dist`.


MattIPv4:
> I added a / to `basePath` since I don't know what to do when the file is directly under the package and not under `dist`.

You can just leave it as an empty string

MattIPv4 (approved):
LGTM

PR #1376, opened 2022-10-27
Update superagent w/ npm auto-update

waaronking:
Newer versions of superagent are found in the dist folder and are currently missing from cdnjs.com


MattIPv4 (approved):
LGTM

PR #1379, opened 2022-10-29
Add materialstyle w/ npm auto-update

nkdas91:


packages/m/materialstyle.json
```diff
@@ -0,0 +1,51 @@
...
+    "front-end",
+    "framework",
+    "web",
+    "bootstrap",
+    "bootstrap-material-design",
+    "bootstrap-theme",
+    "google-material-design"
+  ],
+  "license": "MIT",
+  "name": "@materialstyle/materialstyle",
```
MattIPv4, line 23:
```suggestion
  "name": "materialstyle",
```
This should match the filename

packages/m/materialstyle.json
```diff
@@ -0,0 +1,51 @@
...
+    "url": "https://github.com/materialstyle/materialstyle"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "@materialstyle/materialstyle",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*.@(js|css|map)"
```
MattIPv4, line 35:
```suggestion
          "js/*.js?(.map)",
          "css/*.css?(.map)"
```
I don't think we need a glob-star here (it can lead to files being included that we don't want). The package seems to have all the CSS directly in `dist/css`, and all the JS directly in `dist/js`

MattIPv4 (changes requested):

MattIPv4 (approved):
LGTM

PR #1399, opened 2022-11-16
Update react-router-dom w/ npm auto-update

renbaoshuo:
The location of `react-router-dom.production.min.js` has changed since `react-router-dom@6.4.0`.


MattIPv4 (approved):
LGTM

PR #1405, opened 2022-11-23
Add hexo-theme-keep w/ npm auto-update

XPoet:
**Library name**
hexo-theme-keep

**Library license**
https://github.com/XPoet/hexo-theme-keep/blob/master/LICENSE

**Library source**
https://www.npmjs.com/package/hexo-theme-keep

### DIFF
--- a/packages/h/hexo-theme-keep.json
+++ b/packages/h/hexo-theme-keep.json
@@ -1,0 +1,39 @@
+{
+  "name": "hexo-theme-keep",
+  "description": "A simple and elegant theme for Hexo.",
+  "keywords": [
+    "hexo",
+    "theme",
+    "simple",
+    "elegant",
+    "beautiful",
+    "powerful"
+  ],
+  "license": "GPL-3.0",
+  "homepage": "https://xpoet.cn",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/XPoet/hexo-theme-keep.git"
+  },
+  "authors": [
+    {
+      "name": "XPoet",
+      "email": "i@xpoet.cn",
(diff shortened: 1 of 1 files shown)


packages/h/hexo-theme-keep.json
```diff
@@ -0,0 +1,39 @@
...
+  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "hexo-theme-keep",
+    "fileMap": [
+      {
+        "basePath": "source",
+        "files": [
+          "js/**/*",
+          "fontawesome/**/*",
```
MattIPv4, line 33: Fontawesome already exists on cdnjs, I do not see the value in re-distributing it here

packages/h/hexo-theme-keep.json
```diff
@@ -0,0 +1,39 @@
...
+    }
+  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "hexo-theme-keep",
+    "fileMap": [
+      {
+        "basePath": "source",
+        "files": [
+          "js/**/*",
```
MattIPv4, line 32: We try to avoid using open globstars as it can result in matching files that we don't want. Please can you scope this to just the files that are needed?

packages/h/hexo-theme-keep.json
```diff
@@ -0,0 +1,39 @@
...
+          "fontawesome/**/*",
+          "images/**/*"
```
MattIPv4, line 34: Same here, please avoid using open globstars

MattIPv4 (changes requested):

MattIPv4:
No progress, closing.

PR #1448, opened 2023-01-27
Add docsearch-css w/ npm auto-update

jerryc127:
https://www.npmjs.com/package/@docsearch/css?activeTab=readme

### DIFF
--- a/packages/d/docsearch-css.json
+++ b/packages/d/docsearch-css.json
@@ -19,7 +19,7 @@
       {
         "basePath": "dist",
         "files": [
-          "**/*"
+          "*.css"
         ]
       }
     ]
--- a/packages/d/docsearch-css.json
+++ b/packages/d/docsearch-css.json
@@ -24,8 +24,10 @@
       }
     ]
   },
-"author": {
-    "name": "Algolia, Inc.",
-    "url": "https://www.algolia.com"
-  }
+  "author": [
+    {
+      "name": "Algolia, Inc.",
+      "url": "https://www.algolia.com"
+    }
+  ]
 }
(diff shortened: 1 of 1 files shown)


packages/d/docsearch-css.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "homepage": "https://docsearch.algolia.com/",
+  "autoupdate": {
+    "source": "npm",
+    "target": "@docsearch/css",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 22: 👋 We try to avoid using open globstars, as it can lead to including files we didn't intend to include on the CDN -- it looks like `*.css` would match everything needed here?

MattIPv4 (changes requested):

packages/d/docsearch-css.json
jerryc127: > 👋 We try to avoid using open globstars, as it can lead to including files we didn't intend to include on the CDN -- it looks like `*.css` would match everything needed here?

yes, all css.

MattIPv4: Okay, if this is all CSS, please update the glob pattern to be `*.css` as I had suggested, we don't want open globstars as you currently have.

jerryc127: > Okay, if this is all CSS, please update the glob pattern to be `*.css` as I had suggested, we don't want open globstars as you currently have.

ok, it done

MattIPv4 (approved):
LGTM

PR #1449, opened 2023-01-27
Add docsearch-js w/ npm auto-update

jerryc127:


packages/d/docsearch-js.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "homepage": "https://docsearch.algolia.com/",
+  "autoupdate": {
+    "source": "npm",
+    "target": "@docsearch/js",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 22: :wave: We try to avoid using open globstars in configs as it can lead to including files unintentionally -- looking at the package, you can likely use `@(esm|umd)/*.@(js|d.ts)`

MattIPv4 (changes requested):

packages/d/docsearch-js.json
jerryc127: ok

MattIPv4 (approved):
LGTM

PR #1470, opened 2023-02-23
Add ffmpeg w/ npm auto-update

renbaoshuo:


packages/f/ffmpeg.json
```diff
@@ -0,0 +1,26 @@
...
+  "license": "MIT",
+  "autoupdate": {
+    "source": "npm",
+    "target": "@ffmpeg/ffmpeg",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*.js",
+          "**/*.js.map"
```
MattIPv4, line 17: 👋 We try to avoid using globstars in configs as they can lead to including files that weren't intended to be on the CDN.

Looking at the matched contents, I suspect you can just use this here:

```suggestion
          "*.js?(.map)"
```

MattIPv4 (changes requested):

packages/f/ffmpeg.json
renbaoshuo: In `@ffmpeg/ffmpeg@0.12.0-alpha.1`, they changed the way they packaged it, which caused the file paths to change.

_After:_



_Before:_

MattIPv4: In that case, I would recommend using `{,umd/,esm/}*.js`

renbaoshuo: Okay I've updated it just now.

MattIPv4 (approved):
LGTM

PR #1471, opened 2023-02-23
Add ffmpeg-core w/ npm auto-update

renbaoshuo:


packages/f/ffmpeg-core.json
```diff
@@ -0,0 +1,27 @@
...
+  "autoupdate": {
+    "source": "npm",
+    "target": "@ffmpeg/core",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*.js",
+          "**/*.js.map",
+          "**/*.wasm"
```
MattIPv4, line 18:
```suggestion
          "*.js?(.map)",
          "*.wasm"
```
As with your other PR, we like to avoid using globstars in configs

MattIPv4 (changes requested):

packages/f/ffmpeg-core.json
renbaoshuo: > In `@ffmpeg/ffmpeg@0.12.0-alpha.1`, they changed the way they packaged it, which caused the file paths to change.
>
> _After:_
>
> 
>
> _Before:_
>
> 

-- https://github.com/cdnjs/packages/pull/1470#discussion_r1118169185

MattIPv4: As with #1470, if you need multiple directories, you can use a list: `{,umd/,esm/}*.js`

MattIPv4 (approved):
LGTM

MattIPv4:
Admin merging this -- the CI failure looks to be due to a bug in the CI, the glob pattern matches files if I pull the alpha version down locally. I have added this to the list of bugs in https://github.com/orgs/cdnjs/projects/2/views/1

Chico1244:
A

Issue #1466, opened 2023-02-18
Add: ovh


martinez-hugo:
### Library name

node-ovh

### Library license

https://github.com/ovh/node-ovh/blob/master/LICENSE

### Library source

https://github.com/ovh/node-ovh

### Files to include

- [ovh.es6.js](https://github.com/ovh/node-ovh/blob/master/lib/ovh.es6.js)
- [ovh.es5.js](https://github.com/ovh/node-ovh/blob/master/lib/ovh.es5.js)
- [handler-maker.js](https://github.com/ovh/node-ovh/blob/master/lib/handler-maker.js)
- [endpoints.js](https://github.com/ovh/node-ovh/blob/master/lib/endpoints.js)

martinez-hugo:
PR #1472

MattIPv4:
Closing per discussion in the PR, this library is not built for the web.

PR #1472, opened 2023-02-23
Add ovh w/ npm auto-update

martinez-hugo:


martinez-hugo:
Issue #1466

MattIPv4 (changes requested):
A few change requests here:

1. Your entire file seems to have odd indenting (everything seems to have an extra 2-space indent, including the first line).

2. Your file ending seems to be rather odd, can you check that you have a standard single newline at the end of the file with no trailing whitespace?

3. You indicate you are trying to add the `ovh` package from NPM, but your autoupdate config is currently targeting the `open` package?

4. Inspecting the `ovh` package locally, both the es5 & es6 files in `lib` make use of Node.js `require` calls, which do not work on the web? Is there a different web-compatible bundle that you are intending to add here?

martinez-hugo:
Hi ! Normally I fix all 3 first points

But for 4th point, I've checked repo but there are not other file of this module without require
How can we do ?

MattIPv4:
If you are looking to use this package on the web, you will need to work with the maintainers of the package to have a web-compatible version published, which could then be added to cdnjs.

martinez-hugo:
Does this module is better ?
https://www.npmjs.com/package/@ovh-api/sms

MattIPv4:
No, and you can check this very easily locally rather than asking me...

```shell 
npm i @ovh-api/sms
cd node_modules/@ovh-api/sms
ls # Observe index.js and cjs/
cat index.js # Observe `import` calls
cd cjs
ls # Observe index.js
cat index.js # Observe `require` calls
```

MattIPv4:
No progress, closing.

PR #1480, opened 2023-02-28
Update fancyapps-ui w/ npm auto-update

jerryc127:


packages/f/fancyapps-ui.json
```diff
@@ -34,7 +34,8 @@
       {
         "basePath": "dist",
         "files": [
-          "*.@(js|css)"
+          "*.js",
+          "*/*.@(js|css)"
```
MattIPv4, line 38:
```suggestion
          "*/*.@(js|css)",
          "*/l10n/*.js"
```
This will let you include those l10n files

MattIPv4 (changes requested):
The `"filename": "fancybox.umd.js"` line at the end of the config needs to be updated -- I imagine it either needs to be `index.umd.js` now, or `fancybox/fancybox.umd.js`.

jerryc127:
> The `"filename": "fancybox.umd.js"` line at the end of the config needs to be updated -- I imagine it either needs to be `index.umd.js` now, or `fancybox/fancybox.umd.js`.

OK, I have change

PR #1484, opened 2023-03-05
Add dragselect w/ npm auto-update

### LINKED ISSUE #1483: Add: DragSelect
### Library name

DragSelect

### Library license

https://github.com/ThibaultJanBeyer/DragSelect/blob/master/LICENSE

### Library source

https://www.npmjs.com/package/dragselect

### Files to include

- dist/DragSelect.es6m.js
- dist/DragSelect.js
- dist/ds.es6m.min.js
- dist/ds.min.js

ThibaultJanBeyer:
Resolves #1483

### DIFF
--- a/packages/d/dragselect.json
+++ b/packages/d/dragselect.json
@@ -29,7 +29,7 @@
       {
         "basePath": "dist",
         "files": [
-          "**/*"
+          "*.js"
         ]
       }
     ]
--- a/packages/d/dragselect.json
+++ b/packages/d/dragselect.json
@@ -1,0 +1,43 @@
+{
+  "name": "dragselect",
+  "filename": "ds.min.js",
+  "description": "Easy JavaScript library for selecting and moving elements. With no dependencies. Drag-Select & Drag-And-Drop.",
+  "keywords": [
+    "javascript",
+    "lightweight",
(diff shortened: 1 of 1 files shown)


packages/d/dragselect.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "license": "MIT",
+  "autoupdate": {
+    "source": "npm",
+    "target": "dragselect",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 32: 👋 We try to avoid using open globstars as it can lead to including files on the CDN that weren't intended. Looking at the package, a simple `*.js` would suffice here (if typings are needed, then an additional `**/*.d.ts` may be appropriate, though generally typings aren't needed on a web CDN).

```suggestion
          "*.js"
```

MattIPv4 (changes requested):

packages/d/dragselect.json
ThibaultJanBeyer: Thank you! Let's skip the typings for now then.

MattIPv4 (approved):
LGTM

PR #1488, opened 2023-03-09
Update antd-mobile w/ npm auto-update

ssbabysong:


packages/a/antd-mobile.json
```diff
@@ -24,7 +24,7 @@
     "target": "antd-mobile",
     "fileMap": [
       {
-        "basePath": "dist",
+        "basePath": "bundle",
         "files": [
           "**/!(*.less)"
```
MattIPv4, line 29: Looking at the output from the CI, I think this can be reduced to something far safer now:

```suggestion
          "*.@(js|css)"
```

MattIPv4 (changes requested):
It looks like the filename has slightly changed:

```
  "filename": "antd-mobile.umd.min.js",
```

MattIPv4 (approved):
LGTM

PR #1492, opened 2023-03-11
Add link-styler w/ npm auto-update

sheikhoo:
Add link-styler

### DIFF
--- a/packages/l/link-styler.json
+++ b/packages/l/link-styler.json
@@ -14,8 +14,8 @@
     "mention"
   ],
   "autoupdate": {
-    "source": "git",
-    "target": "git://github.com/sheikhoo/link-styler",
+    "source": "npm",
+    "target": "link-styler",
     "fileMap": [
       {
         "basePath": "dist",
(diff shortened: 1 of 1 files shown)


packages/l/link-styler.json
```diff
@@ -0,0 +1,34 @@
...
+  "license": "MIT",
+  "homepage": "https://link-styler.js.org",
+  "keywords": [
+    "autolink",
+    "hashtag",
+    "mention"
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/sheikhoo/link-styler",
```
MattIPv4, line 18: Our auto-update mechanism requires that either there exists a source repository with versions tagged that contain the required files for the CDN, or a package released on NPM that contains the files needed for the CDN.

It looks like you have a versioned NPM package, so switching to using that might be best?

MattIPv4 (changes requested):

packages/l/link-styler.json
sheikhoo: Thank you, I changed git to npm

MattIPv4 (approved):
LGTM

PR #1552, opened 2023-06-01
Add primeflex w/ npm auto-update

darkvertex:
Saw you have primevue and primeicons but not their brother https://github.com/primefaces/primeflex


MattIPv4 (approved):
LGTM

PR #1570, opened 2023-06-19
Add jsonpath w/ npm auto-update

ranlima:


packages/j/jsonpath-dchester.json
MattIPv4, line 1: I would probably expect this file to be named `jsonpath.json`, to match the project name

packages/j/jsonpath-dchester.json
```diff
@@ -0,0 +1,42 @@
...
+    "type": "git",
+    "url": "https://github.com/dchester/jsonpath.git"
+  },
+  "license": "MIT",
+  "autoupdate": {
+    "source": "npm",
+    "target": "jsonpath",
+    "fileMap": [
+      {
+        "basePath": ".",
```
MattIPv4, line 28:
```suggestion
        "basePath": "",
```
If we need the root, the `basePath` can be empty

packages/j/jsonpath-dchester.json
```diff
@@ -0,0 +1,42 @@
...
+        "files": [
+          "**/jsonpath.js",
+          "**/jsonpath.min.js"
```
MattIPv4, line 31:
```suggestion
          "jsonpath?(.min).js"
```
We try to avoid using globstars wherever possible, as it can lead to including unexpected files in the CDN. It looks like these files are at the root of the package, so we don't need to specify any directory/globstar. (We can also simplify this into a single glob)

MattIPv4 (changes requested):

packages/j/jsonpath-dchester.json
ranlima: Changed

packages/j/jsonpath-dchester.json
ranlima: Changed

packages/j/jsonpath-dchester.json
ranlima: Changed

MattIPv4 (approved):
LGTM

Issue #1629, opened 2023-08-22
Add: plotly/dash


yzcheung:
### Library name

plotly/dash

### Library license

https://github.com/plotly/dash/blob/dev/LICENSE

### Library source

https://www.jsdelivr.com/package/npm/dash-renderer
https://www.jsdelivr.com/package/npm/dash-core-components
https://www.jsdelivr.com/package/npm/dash-html-components
https://www.jsdelivr.com/package/npm/dash-table

### Files to include

[/dash-renderer/build/dash_renderer.min.js](https://cdn.jsdelivr.net/npm/dash-renderer@latest/build/dash_renderer.min.js)
[/dash_core_components/dash_core_components.min.js](https://cdn.jsdelivr.net/npm/dash-core-components@latest/dash_core_components/dash_core_components.js)
[/dash_core_components/dash_core_components-shared.js](https://cdn.jsdelivr.net/npm/dash-core-components@latest/dash_core_components/dash_core_components-shared.js)
[/dash_html_components/dash_html_components.min.js](https://cdn.jsdelivr.net/npm/dash-html-components@latest/dash_html_components/dash_html_components.min.js)
[/dash_table/bundle.js](https://cdn.jsdelivr.net/npm/dash-table@latest/dash_table/bundle.js)

PR #1642, opened 2023-09-04
Add dash-renderer w/ npm auto-update

MattIPv4:
Adding dash-renderer using npm auto-update from dash-renderer.

Resolves (part of) #1629.


MattIPv4 (commented):
lgtm

PR #1643, opened 2023-09-04
Add dash-core-components w/ npm auto-update

MattIPv4:
Adding dash-core-components using npm auto-update from dash-core-components.

Resolves (part of) #1629.


MattIPv4 (commented):
lgtm

PR #1644, opened 2023-09-04
Add dash-html-components w/ npm auto-update

MattIPv4:
Adding dash-html-components using npm auto-update from dash-html-components.

Resolves (part of) #1629.


MattIPv4 (commented):
lgtm

PR #1645, opened 2023-09-04
Add dash-table w/ npm auto-update

MattIPv4:
Adding dash-table using npm auto-update from dash-table.

Resolves (part of) #1629.


MattIPv4 (commented):
lgtm

PR #1683, opened 2023-10-19
Add dj-lightbox w/ git auto-update

danyg790:


packages/d/dj-lightbox.json
```diff
@@ -0,0 +1,36 @@
...
+    "images"
+  ],
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/danyg790/dj-lightbox.git",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 21: Please avoid using glob stars, as these can lead to files being included in the CDN that were not intended. You should ideally be specifying wildcards with extensions, for example `*.js`, or `*.@(css|js)`

MattIPv4 (changes requested):

packages/d/dj-lightbox.json
danyg790: I have made the change to eliminate the glob stars and set it to include only the js and css files

danyg790:
I have made the change to eliminate the glob stars and set it to include only the js and css files

danyg790:
Why am I getting the "Error: no version found on git", and how would I fix that?

Issue #1685, opened 2023-10-25
Update: layui


sentsim:
### Library name

layui

### Library JSON file

https://github.com/cdnjs/packages/blob/master/packages/l/layui.json

### Library source

https://github.com/layui/layui

### Files to include

- "*.map"

PR #1686, opened 2023-10-25
Update layui w/ git auto-update

sentsim:
- updated homepage
- updated the files to include  ( resolves: #1685 )
- updated author name and delete author's email (out of use)

### DIFF
--- a/packages/l/layui.json
+++ b/packages/l/layui.json
@@ -7,7 +7,7 @@
     "type": "git",
     "url": "git+ssh://git@github.com/layui/layui.git"
   },
-  "homepage": "https://www.layui.com",
+  "homepage": "https://layui.dev",
   "keywords": [
     "layui",
     "ui",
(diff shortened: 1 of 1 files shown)


packages/l/layui.json
```diff
@@ -24,15 +24,15 @@
         "files": [
           "*.js",
           "css/**/*.css",
-          "font/*.@(eot|ttf|svg|woff|woff2)"
+          "font/*.@(eot|ttf|svg|woff|woff2)",
+          "*.map"
```
MattIPv4, line 28:
```suggestion
          "*.js?(.map)",
          "css/**/*.css",
          "font/*.@(eot|ttf|svg|woff|woff2)"
```
Instead of including all `map` files, I recommend editing the existing `*.js` line to also allow sourcemaps, `*.js?(.map)`

MattIPv4 (changes requested):

MattIPv4 (approved):
LGTM, ty

PR #1688, opened 2023-10-26
Add graphql-yoga w/ git auto-update

### LINKED ISSUE #1687: Add: graphql-yoga
### Library name

graphql-yoga

### Library license

https://github.com/dotansimha/graphql-yoga/blob/main/LICENSE

### Library source

https://www.npmjs.com/package/graphql-yoga

### Files to include

- `dist/**`

YassinEldeeb:
closes #1687

### DIFF
--- a/packages/g/graphql-yoga.json
+++ b/packages/g/graphql-yoga.json
@@ -1,0 +1,44 @@
+{
+  "name": "graphql-yoga",
+  "description": "Rewrite of a fully-featured GraphQL Server with focus on easy setup, performance & great developer experience. The core of Yoga implements WHATWG Fetch API and can run/deploy on any JS environment.",
+  "keywords": [
+    "nodejs",
+    "javascript",
+    "fetch",
+    "graphql",
+    "typescript",
+    "w3c",
+    "graphql-server",
+    "whatwg",
+    "bun",
+    "deno",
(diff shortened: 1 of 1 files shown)


packages/g/graphql-yoga.json
```diff
@@ -0,0 +1,44 @@
...
+      "email": "saihajpreet.singh@gmail.com"
+    }
+  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "graphql-yoga",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": ["*.js"]
```
MattIPv4, line 39: Looking at the `graphql-yoga` package on [NPM](https://www.npmjs.com/package/graphql-yoga?activeTab=code), there is no `dist` directory.

It looks like here are `cjs` and `esm` directories that contain code, but the `cjs` directory makes use of Node.js `require` calls that will not work in a browser, and the `esm` directory imports external modules which won't work unless module mapping is used afaik.

Is there a pre-bundled version that you should be attempting to reference here instead? Is this library meant to be used in a browser -- it looks like a server framework?

MattIPv4 (changes requested):

MattIPv4:
No progress, closing.

PR #1693, opened 2023-11-03
Add jos-animation w/ npm auto-update

jesvijonathan:
Adds [JOS-Animation](https://jos-animation.vercel.app/).

### DIFF
--- a/packages/j/jos-animation.json
+++ b/packages/j/jos-animation.json
@@ -2,7 +2,7 @@
   "name": "jos-animation",
   "description": "An animation library package to instantly add beautiful/professional looking animation to your website",
   "homepage": "jos-animation.vercel.app",
-  "filename": "dist/jos.js",
+  "filename": "jos.js",
   "keywords": [
     "jos",
     "on-scroll",
@@ -35,7 +35,9 @@
     "fileMap": [
       {
         "basePath": "dist",
-        "files": ["**/*"]
+        "files": [
+          "*.@(css|js)"
+        ]
       }
     ]
   },
--- a/packages/j/jos-animation.json
+++ b/packages/j/jos-animation.json
@@ -15,7 +15,6 @@
     "javascript",
     "lightweight",
     "performance",
-    "scroll",
     "event",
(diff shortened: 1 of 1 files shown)


packages/j/jos-animation.json
```diff
@@ -0,0 +1,43 @@
...
+    "type": "git",
+    "url": "git://github.com/jesvijonathan/JOS-Animation-Library.git"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "jos-animation",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": ["**/*"]
```
MattIPv4, line 38:
```suggestion
        "files": [
          "*.@(css|js)"
        ]
```
Please avoid using open globstars, as they can lead to including files not intended for the CDN. I suspect you just need the CSS/JS files directly in the `dist` directory?

(Also, new line for each item in array, allows for cleaner diffs in the future)

packages/j/jos-animation.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "jos-animation",
+  "description": "An animation library package to instantly add beautiful/professional looking animation to your website",
+  "homepage": "jos-animation.vercel.app",
+  "filename": "dist/jos.js",
```
MattIPv4, line 5:
```suggestion
  "filename": "jos.js",
```
As the `basePath` is set to `dist`, we'll pull the files out of that directory into the root.

MattIPv4 (changes requested):

packages/j/jos-animation.json
jesvijonathan: Gotchya

MattIPv4 (approved):
LGTM

jesvijonathan:
@MattIPv4, any idea why the latest of [JOS](https://www.npmjs.com/package/jos-animation) from npm is not appearing in [CDNJS](https://cdnjs.com/libraries/jos-animation) ?

also the version options are limited...

PR #1694, opened 2023-11-10
Update mathjax w/ npm auto-update

### LINKED ISSUE #1669: Update: Mathjax v4.0.0-beta.3 & v4.0.0-beta.4
### Library name

Mathjax

### Library JSON file

https://github.com/cdnjs/packages/blob/master/packages/m/mathjax.json

### Library source

https://www.npmjs.com/package/mathjax

### Files to include

- **/!(node-main.js)

Pinghigh:
Should fix #1669.

### DIFF
--- a/packages/m/mathjax.json
+++ b/packages/m/mathjax.json
@@ -21,6 +21,13 @@
         "basePath": "",
         "files": [
           "es5/**/!(node-main.js)",
+          "*.js!(node-main.js)",
+          "ui/*.js",
+          "sre/*.js",
+          "adaptors/*.js",
+          "a11y/*.js",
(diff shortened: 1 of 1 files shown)


Pinghigh:
MathJax 4.0.0-beta.3~4 does not have `es5/` folder but the version before 4.0.0-alpha.1 has it. That caused the error.

Should I delete it?

MattIPv4:
Sorry for the delay in getting back to you. Yeah, if it no longer exists in new versions, it doesn't really make sense to keep it in the config. You'll want to provide a new `filename` though, for whatever the recommended default file to use should be now.

PR #1700, opened 2023-11-20
Update layui w/ git auto-update

sentsim:
The current files matching rule does not include `layui.css.map` (https://cdnjs.com/libraries/layui/2.9.0-beta.1).
All files in the `dist` directory should be included.
ref: https://unpkg.com/browse/layui@2.9.0-beta.1/dist/

### DIFF
--- a/packages/l/layui.json
+++ b/packages/l/layui.json
@@ -22,7 +22,9 @@
       {
         "basePath": "dist",
         "files": [
-          "**/*"
+          "*.js?(.map)",
+          "css/**/*.css(.map)",
+          "font/*.@(eot|ttf|svg|woff|woff2)"
         ]
       }
     ]
--- a/packages/l/layui.json
+++ b/packages/l/layui.json
@@ -22,9 +22,7 @@
       {
         "basePath": "dist",
         "files": [
-          "*.js?(.map)",
-          "css/**/*.css",
-          "font/*.@(eot|ttf|svg|woff|woff2)"
+          "**/*"
         ]
       }
     ]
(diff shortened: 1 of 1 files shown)


packages/l/layui.json
```diff
@@ -22,9 +22,7 @@
       {
         "basePath": "dist",
         "files": [
-          "*.js?(.map)",
-          "css/**/*.css",
-          "font/*.@(eot|ttf|svg|woff|woff2)"
+          "**/*"
```
MattIPv4, line 25: Generally speaking, we do not allow folks to add open glob-stars in new PRs, as it can lead to files be included that shouldn't be on the CDN. if you're just looking for CSS source maps to be added, please just add that to the existing config, instead of blindly ripping it all out and replacing it with a glob-star.

```suggestion
          "*.js?(.map)",
          "css/**/*.css(.map)",
          "font/*.@(eot|ttf|svg|woff|woff2)"
```

MattIPv4 (changes requested):

packages/l/layui.json
```diff
@@ -23,7 +23,7 @@
         "basePath": "dist",
         "files": [
           "*.js?(.map)",
-          "css/**/*.css",
+          "css/**/*.css(.map)",
```
MattIPv4, line 26:
```suggestion
          "css/**/*.css?(.map)",
```

MattIPv4 (approved):

MattIPv4:
I don't really see any reason to make such a change?

sentsim:
Okay, Then let's keep it as is. Thanks again.  👍

PR #1722, opened 2023-12-05
Add better-ajv-errors w/ npm auto-update

### LINKED ISSUE #1721: Add: better-ajv-errors
### Library name

better-ajv-errors

### Library license

https://github.com/atlassian/better-ajv-errors/blob/main/LICENSE

### Library source

https://www.npmjs.com/package/better-ajv-errors

### Files to include

lib/cjs/index.js
lib/cjs/validation-errors/*.js
lib/cjs/json/*.js

jeremyfiel:
closes #1721

### DIFF
--- a/packages/b/better-ajv-errors.json
+++ b/packages/b/better-ajv-errors.json
@@ -1,0 +1,36 @@
+{
+  "name": "better-ajv-errors",
+  "description": "JSON Schema validation for Human",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/atlassian/better-ajv-errors.git"
+  },
+  "keywords": [
+    "JSON",
+    "schema",
+    "validator",
+    "jsonschema",
+    "json-schema",
+    "ajv",
+    "ajv-errors"
+  ],
+  "license": "Apache-2.0",
+  "filename": "index.mjs",
+  "autoupdate": {
+    "source": "npm",
+    "target": "better-ajv-errors",
+    "fileMap": [
+      {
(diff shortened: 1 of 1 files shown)


MattIPv4 (changes requested):
🤔 It looks like you're including the cjs version here that uses Node.js `require` calls that won't work in a browser? Potentially we could add the ESM version for folks that're using modules, but ideally there'd be a pre-bundled browser version?

jeremyfiel:
Ya, all of these packages I added as a user.  So I don't have  any way to add the bundling.

MattIPv4:
If this is switched over to the ESM version I'd be happy to add it for folks that're using browser modules, or consuming this through runtimes like Deno etc.

jeremyfiel:
Ok will update it

jeremyfiel:
ok, i think that glob pattern will work.  thanks

packages/b/better-ajv-errors.json
```diff
@@ -0,0 +1,36 @@
...
+  "license": "Apache-2.0",
+  "filename": "index.mjs",
+  "autoupdate": {
+    "source": "npm",
+    "target": "better-ajv-errors",
+    "fileMap": [
+      {
+        "basePath": "lib/esm",
+        "files": [
+          "/**/*.@(mjs?(.map))"
```
MattIPv4, line 26:
```suggestion
          "**/*.mjs?(.map)"
```

MattIPv4 (approved):

PR #1732, opened 2023-12-26
Add carbon-charts w/ npm auto-update

nstuyvesant:


nstuyvesant (commented):
Sorry I left out the extension.

packages/c/carbon-charts.json
```diff
@@ -0,0 +1,63 @@
...
+  "filename": "dist/umd/bundle.umd.js",
+  "autoupdate": {
+    "source": "npm",
+    "target": "@carbon/charts",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*.css",
+          "**/*.css.map"
```
MattIPv4, line 47:
```suggestion
          "*.css?(.map)"
```
You can combine this into a single glob, and I don't believe you need a glob star here, as the CSS files aren't in nested folders?

packages/c/carbon-charts.json
```diff
@@ -0,0 +1,63 @@
...
+        ]
+      },
+      {
+        "basePath": "dist/umd",
+        "files": [
+          "**/*.@(js|map)"
```
MattIPv4, line 53:
```suggestion
          "*.js?(.map)"
```
Same deal here?

packages/c/carbon-charts.json
```diff
@@ -0,0 +1,63 @@
...
+		"data",
+		"typescript"
+  ],
+  "license": "Apache-2.0",
+  "homepage": "https://www.carbondesignsystem.com/",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/carbon-design-system/carbon-charts/tree/master/packages/core"
+  },
+  "filename": "dist/umd/bundle.umd.js",
```
MattIPv4, line 38:
```suggestion
  "filename": "bundle.umd.js",
```
As you're setting `basePath`, we'll pull files up out of that directory

MattIPv4 (changes requested):

MattIPv4 (approved):
LGTM

PR #1736, opened 2023-12-30
Add graphics-element w/ npm auto-update

Pomax:
JSON set up to autoupdate from `npm`, using https://www.npmjs.com/package/graphics-element


Pomax:
The stars and package downloads make sense, given that I only just released it. I'm switching https://pomax.github.io/bezierinfo, https://pomax.github.io/bezierjs, and https://pomax.github.io/are-we-flying over to it.

MattIPv4 (approved):

Issue #1737, opened 2024-01-01
Add: view-ui-plus


KazmaZhang:
### Library name

view-ui-plus

### Library license

https://github.com/view-design/ViewUIPlus/blob/master/LICENSE

### Library source

https://www.npmjs.com/package/view-ui-plus

### Files to include

dist/*

PR #1738, opened 2024-01-02
Add view-ui-plus w/ npm auto-update

KazmaZhang:
Closes: #1737

### DIFF
--- a/packages/v/view-ui-plus.json
+++ b/packages/v/view-ui-plus.json
@@ -1,0 +1,40 @@
+{
+  "name": "view-ui-plus",
+  "description": "An enterprise-level UI component library and front-end solution based on Vue.js 3",
+  "homepage": "www.iviewui.com",
+  "keywords": [
+    "iview",
+    "vue",
+    "viewui",
+    "viewuiplus",
+    "vue.js",
+    "component",
+    "components",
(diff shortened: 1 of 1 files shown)


packages/v/view-ui-plus.json
```diff
@@ -0,0 +1,40 @@
...
+    "url": "git+https://github.com/view-design/ViewUIPlus.git"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "view-ui-plus",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 34: Please avoid using (open) globstars, as it can lead to files being included on the CDN that aren't needed, or that we don't support.

Looking at the package, I suspect you want `{,locale/}*.js?(.map)`, `styles/*.css`, and `styles/fonts/*.@(svg|ttf|woff|woff2)`?

MattIPv4 (changes requested):

packages/v/view-ui-plus.json
KazmaZhang: Updated, thx u.

MattIPv4 (approved):
LGTM

PR #1739, opened 2024-01-08
Update jodit w/ npm auto-update

xdan:
Hi, I have new build structure https://www.npmjs.com/package/jodit?activeTab=code


xdan:
Hi, @KazmaZhang can you check it?

xdan:
@MattIPv4 Hi, can you check it?

MattIPv4 (commented):
👀 I'm not sure this is really the config you want. When you specify a `basePath`, we pull assets _out_ of that folder into the root. So if you have `es2021/index.js` and `es2018/index.js`, then we're going to randomly pick one of those to be `index.js` essentially.

If you want to include multiple directories like this, you probably just want a single entry with an empty `basePath`.

xdan:
@MattIPv4 Hi, again. I already tried an empty basePath, it crashes https://github.com/cdnjs/packages/actions/runs/7525599621/job/20482198820?pr=1739
Maybe because there are a lot of files. Maybe I need to reduce the number of files?

MattIPv4:
Yes, there does seem to be a very very large number of files being matched due to the globstar in the config.

Could I suggest setting the basePath to es5 (or es2015/es2012/es2021) and just including those files, as there is far less of them and they should work in all browsers? Your esm code includes a mix of relative and package imports, so I'm not sure that'd even work in a browser.

xdan:
@MattIPv4 Hi again, I fixed it.

xdan:
@MattIPv4 Would you mind taking a look at this, please?

packages/j/jodit.json
```diff
@@ -19,10 +19,9 @@
     "target": "jodit",
     "fileMap": [
       {
-        "basePath": "build",
+        "basePath": "",
         "files": [
-          "jodit*.+(css|js)",
-          "plugins/**/*.@(css|js)"
+          "!(examples|types)/**/*min.@(css|js)"
```
MattIPv4, line 24:
```suggestion
          "@(es2015|es2018|es2021|es2021.en|es5)/**/*.min.@(css|js)"
```
I would be much more comfortable with leaving the globstar in here if we specify the specific directories we want to pull from, rather than saying "everything everywhere except examples/types" which would leave us far more at risk of unexpected things being included down the road

MattIPv4 (approved):
LGTM

xdan:
@MattIPv4Hi, when will this fix be rolled out? I still don't have the files https://cdnjs.com/libraries/jodit

MattIPv4:
No new version has been released since this was merged; https://github.com/cdnjs/logs/tree/prod/packages/j/jodit

PR #1762, opened 2024-02-23
Update mathjax w/ npm auto-update

### LINKED ISSUE #1669: Update: Mathjax v4.0.0-beta.3 & v4.0.0-beta.4

Pinghigh:
Should fix #1669, #1694

As described in #1694 
>MathJax 4.0.0-beta.3~4 does not have es5/ folder but the version before 4.0.0-alpha.1 has it. That caused the error.
I changed the config with `es5` directory in this PR.

But 4.x is still a beta version and there is no stable 4.x release yet, so I'm not sure we should do the update now.
Please merge as appropriate, thanks

### DIFF
--- a/packages/m/mathjax.json
+++ b/packages/m/mathjax.json
@@ -20,7 +20,6 @@
       {
         "basePath": "",
         "files": [
-          "**/!(node-main.js)",
           "*.js!(node-main.js)",
           "ui/*.js",
           "sre/*.js",
--- a/packages/m/mathjax.json
+++ b/packages/m/mathjax.json
@@ -21,6 +21,13 @@
         "basePath": "",
         "files": [
           "es5/**/!(node-main.js)",
+          "*.js!(node-main.js)",
+          "ui/*.js",
+          "sre/*.js",
+          "adaptors/*.js",
+          "a11y/*.js",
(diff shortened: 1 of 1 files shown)


packages/m/mathjax.json
```diff
@@ -20,7 +20,14 @@
       {
         "basePath": "",
         "files": [
-          "es5/**/!(node-main.js)",
+          "**/!(node-main.js)",
```
MattIPv4, line 23: We tend to avoid globstars, and especially open ones -- this line is now basically saying "add every file ever, except node-main.js", which I suspect is not what you want?

MattIPv4 (changes requested):

packages/m/mathjax.json
Pinghigh: 3.x.x has the `es5` directory, if we do not need to think about it, we can apply this change, thanks for reply.

MattIPv4 (approved):
LGTM

PR #1783, opened 2024-03-26
Create swiper-bundle.min.css

lekanshenghuo:


MattIPv4:
Your PR should be adding a JSON configuration file for a package. We will handle then pulling in the actual files for the library based on the configuration file. You should not be adding library files yourself.

PR #1825, opened 2024-06-12
Add geostats w/ git auto-update

fcorallini:


MattIPv4 (changes requested):
As the CI error notes, your package JSON config file requires an `autoupdate` definition, so we know where to fetch the versions from and what files to include.

fcorallini:
> As the CI error notes, your package JSON config file requires an `autoupdate` definition, so we know where to fetch the versions from and what files to include.

#3b5dc56

MattIPv4 (approved):
LGTM, ty

PR #1826, opened 2024-06-18
Update carbon-charts w/ npm auto-update

nstuyvesant:
## Change
- Carbon Charts started adding .cjs extension in place of .js for UMD files as the package.json's **type** was set to **module** causing Webpack to misidentify the UMD bundles as ESM.
- New naming follows the default that Vite.js uses.


MattIPv4 (approved):
LGTM, ty

PR #1827, opened 2024-06-21
Add erc725js w/ npm auto-update

CJ42:
https://www.npmjs.com/package/@erc725/erc725.js

### DIFF
--- a/packages/e/erc725js.json
+++ b/packages/e/erc725js.json
@@ -1,0 +1,55 @@
+{
+  "name": "erc725.js",
+  "filename": "erc725.min.js",
+  "description": "Library to interact with ERC725 smart contracts and encode/decode ERC725Y data keys.",
+  "homepage": "https://docs.lukso.tech/tools/erc725js/getting-started",
+  "keywords": [
+    "ethereum", 
+    "erc725", 
+    "lsp",
+    "lukso"
+  ],
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ERC725Alliance/erc725.js/tree/develop"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "@erc725/erc725.js",
+    "fileMap": [
+      {
+        "basePath": "",
+        "files": [
+          "**/*.@(js|ts)"
+        ]
+      }
+    ]
+  },
+  "authors": [
+    {
+      "name": "Robert McLeod",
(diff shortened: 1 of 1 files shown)


magalimorin18 (approved):
good idea

packages/e/erc725js.json
```diff
@@ -0,0 +1,55 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ERC725Alliance/erc725.js/tree/develop"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "@erc725/erc725.js",
+    "fileMap": [
+      {
+        "basePath": "",
```
MattIPv4, line 22:
```suggestion
        "basePath": "build/main/src",
```
The files under `build/main/src` make use of Node.js `require` calls and will not work in a browser. As such, I suspect you just want the files from `build/module/src`, which should work in browsers that support ESM.

packages/e/erc725js.json
```diff
@@ -0,0 +1,55 @@
...
+        "files": [
+          "**/*.@(js|ts)"
```
MattIPv4, line 24:
```suggestion
          "**/*.@(js|d.ts)"
```

packages/e/erc725js.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "erc725.js",
+  "filename": "erc725.min.js",
```
MattIPv4, line 3:
```suggestion
  "filename": "index.min.js",
```
It looks like the entrypoint file for the library is `index.js` (or `index.min.js` that we'll generate), not `erc725.min.js`

MattIPv4 (changes requested):

packages/e/erc725js.json
```diff
@@ -0,0 +1,55 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ERC725Alliance/erc725.js/tree/develop"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "@erc725/erc725.js",
+    "fileMap": [
+      {
+        "basePath": "build/main/src",
```
MattIPv4, line 22:
```suggestion
        "basePath": "build/module/src",
```
The `main` build appears to make use of Node.js `require` calls that will not work in a browser. The `module` version appears to use `import` statements which should work for folks using ESM in the browser.

MattIPv4 (approved):
LGTM

PR #1830, opened 2024-06-26
Add surreal w/ git auto-update

jph00:
Adding Surreal.js library to cdnjs. 

- Library: https://github.com/gnat/surreal/
- Auto-update configured for git
- Files to be included: surreal.js

### DIFF
--- a/packages/s/surreal.json
+++ b/packages/s/surreal.json
@@ -1,6 +1,6 @@
 {
   "name": "surreal",
-  "filename": "surreal.min.js",
+  "filename": "surreal.js",
   "description": "A lightweight JavaScript library for creating and manipulating HTML in a fluent, jQuery-like way.",
   "keywords": [
     "javascript",
--- a/packages/s/surreal.json
+++ b/packages/s/surreal.json
@@ -28,7 +28,7 @@
   },
   "authors": [
     {
-      "name": "Nathanael Anderson"
+      "name": "Nathaniel Sabanski"
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


jph00:
Thanks @MattIPv4 -- sorry about that. Should be fixed now.

packages/s/surreal.json
```diff
@@ -0,0 +1,34 @@
...
+  ],
+  "homepage": "https://github.com/gnat/surreal",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/gnat/surreal.git"
+  },
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/answerdotai/surreal.git",
```
MattIPv4, line 19: I'm not sure I'm a fan of using a fork of the source repository as the auto-update target here, as there is no guarantee the source code isn't being tampered with (I even see a commit in the repo that does modify the source). Please work with the original author to have it tagged or published to NPM instead.

MattIPv4 (changes requested):

jph00:
I've checked with @gnat, and he feels that the approach of pulling from my fork is preferred. He does not want to tag releases in the main project, in order to avoid implicitly encouraging the use of build machinery.

MattIPv4:
Given the context in https://github.com/cdnjs/packages/pull/1831#issuecomment-2267849805, I'm happy to let this through with the different source.

MattIPv4 (approved):
LGTM

PR #1831, opened 2024-06-26
Add css-scope-inline w/ git auto-update

jph00:
Adding CSS Scope Inline library to cdnjs. 

- Library: https://github.com/gnat/css-scope-inline
- Auto-update configured for git
- Files to be included: script.js

### DIFF
--- a/packages/c/css-scope-inline.json
+++ b/packages/c/css-scope-inline.json
@@ -16,7 +16,7 @@
   },
   "autoupdate": {
     "source": "git",
-    "target": "git://github.com/gnat/css-scope-inline.git",
+    "target": "git://github.com/answerdotai/css-scope-inline.git",
     "fileMap": [
       {
         "basePath": "",
--- a/packages/c/css-scope-inline.json
+++ b/packages/c/css-scope-inline.json
@@ -28,7 +28,7 @@
   },
   "authors": [
     {
-      "name": "Nathanael Anderson"
+      "name": "Nathaniel Sabanski"
     }
   ]
 }
--- a/packages/c/css-scope-inline.json
+++ b/packages/c/css-scope-inline.json
@@ -1,5 +1,6 @@
 {
   "name": "css-scope-inline",
+  "filename": "script.min.js",
   "description": "Lightweight CSS scoping for inline styles without a build step",
   "keywords": [
     "css",
@@ -7,9 +8,8 @@
     "inline",
     "styling"
   ],
-  "author": "Nathanael Anderson",
-  "license": "MIT",
   "homepage": "https://github.com/gnat/css-scope-inline",
+  "license": "MIT",
   "repository": {
     "type": "git",
     "url": "git://github.com/gnat/css-scope-inline.git"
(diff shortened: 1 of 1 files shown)


MattIPv4 (changes requested):
As the CI error indicates, to pull from a GitHub repository, there needs to be tags for each version that we should pull into cdnjs.

jph00:
This should be fixed now too.

packages/c/css-scope-inline.json
```diff
@@ -0,0 +1,34 @@
...
+  ],
+  "homepage": "https://github.com/gnat/css-scope-inline",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/gnat/css-scope-inline.git"
+  },
+  "autoupdate": {
+    "source": "git",
+    "target": "git://github.com/answerdotai/css-scope-inline.git",
```
MattIPv4, line 19: As with #1830, I am not comfortable using a fork of the source repository as the auto-update target here, as there is no guarantee the source code isn't being tampered with. Please work with the original author to have it tagged or published to NPM instead.

MattIPv4 (changes requested):

jph00:
Understood @MattIPv4 -- I reached out to the author once already since I also felt that would be preferred, but didn't get a response. I've actually set up an automatic release github workflow on the fork which pulls directly from the source repo. AFAICT there's no particular reason to believe the source repo would be less likely to have sketchy stuff going on than my version!

If I'm not able to get a response from the original author, what could I do to create a new project based on the original which you'd be comfortable adding to cdnjs? Would I need to give it a new name? In general I assume there's a way for project forks to be added to cdnjs -- quite often forks become more popular than the originals?

gnat:
Hey @MattIPv4 , I'm the author of both of these libraries.

In this specific case, I feel the fork is OK for cdnjs, because:

* I do not plan to add tagged releases to either project, specifically because I do not want to railroad new developers into thinking they need to use build machinery.
  * One of the main goals of these projects is to service new developers, to be real legit examples of libraries you can just download or copy/paste to work.
* Both projects currently have CDN hosted on https://www.jsdelivr.com
  * I explicitly chose jsdelivr because no tagging is required.
  * Unless there's a downside to jsdelivr I'm not considering??
* @jph00 requires the use of cdnjs because of a commercial product that only supports cdnjs.
  * I do not plan to add build machinery just to accommodate the commercial entity- they could add jsdelivr tomorrow and this conversation is moot.

Part of the whole idea behind surreal, css-scope-inline (and IMHO, htmx) is they are useful projects you can just paste the script in and have it work without new developers feeling like they are being railroaded into a toolchain pipeline, which often distracts and clouds their understanding and make them a captive audience to that machinery.

Apologies for dragging you into this. I understand we all have limited bandwidth to make workarounds for private companies.

MattIPv4:
Understood, thank you for the context! Given that, I'll let this go through with the different source repo.

MattIPv4 (approved):
LGTM

jph00:
@MattIPv4 thanks for your help. Question: surreal.js is showing up in the cdn, but this one (css-scope-inline) isn't-- does it just take a while?

jph00:
Apologies, I figured out the source of the problem - pebkac... Made a PR to fix.

PR #1846, opened 2024-07-12
Update element-plus w/ npm auto-update

huankong233:
element-plus has a dark mode

but you need to import this css('element-plus/theme-chalk/dark/css-vars.css')

but there is no such css file in cdnjs

check out here: https://element-plus.org/en-US/guide/dark-mode.html


MattIPv4 (approved):
LGTM, ty

PR #1860, opened 2024-07-31
Add zonos-elements w/ npm auto-update 

Thienhuynh95:
Add zonos-elements to cdnjs

- npm: https://www.npmjs.com/package/@zonos/elements
- auto update configured for npm

### DIFF
--- a/packages/z/zonos-elements.json
+++ b/packages/z/zonos-elements.json
@@ -17,12 +17,6 @@
     "loader/"
   ],
   "license": "MIT",
-  "optimization": {
-    "js": false,
-    "css": false,
-    "png": true,
-    "jpg": true
-  },
   "autoupdate": {
     "source": "npm",
     "target": "zonos-elements",
--- a/packages/z/zonos-elements.json
+++ b/packages/z/zonos-elements.json
@@ -2,13 +2,6 @@
   "name": "zonos-elements",
   "filename": "dist/scripts/loadZonos.js",
   "description": "Zonos component UI for Stripe elements",
-  "main": "dist/index.cjs.js",
-  "module": "dist/index.js",
-  "es2015": "dist/esm/index.mjs",
-  "es2017": "dist/esm/index.mjs",
-  "types": "dist/types/index.d.ts",
-  "unpkg": "dist/zonos-elements/zonos-elements.esm.js",
-  "collection": "dist/collection/collection-manifest.json",
   "keywords": [
     "cross-border",
     "international shopping",
@@ -22,10 +15,6 @@
     "type": "git",
     "url": "https://github.com/Zonos/zonos-elements-readme"
   },
-  "files": [
-    "dist/",
-    "loader/"
-  ],
   "license": "MIT",
   "autoupdate": {
     "source": "npm",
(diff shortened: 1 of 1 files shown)


packages/z/zonos-elements.json
```diff
@@ -0,0 +1,42 @@
...
+  "name": "zonos-elements",
+  "filename": "dist/scripts/loadZonos.js",
+  "description": "Zonos component UI for Stripe elements",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.js",
+  "es2015": "dist/esm/index.mjs",
+  "es2017": "dist/esm/index.mjs",
+  "types": "dist/types/index.d.ts",
+  "unpkg": "dist/zonos-elements/zonos-elements.esm.js",
+  "collection": "dist/collection/collection-manifest.json",
```
MattIPv4, line 11:
```suggestion
```
Please remove these properties, as cdnjs does not support/expect them in a library JSON config.

packages/z/zonos-elements.json
```diff
@@ -0,0 +1,42 @@
...
+    "checkout"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Zonos/zonos-elements-readme"
+  },
+  "files": [
+    "dist/",
+    "loader/"
+  ],
```
MattIPv4, line 28:
```suggestion
```

packages/z/zonos-elements.json
```diff
@@ -0,0 +1,42 @@
...
+  "license": "MIT",
+  "autoupdate": {
+    "source": "npm",
+    "target": "@zonos/elements",
+    "fileMap": [
+      {
+        "basePath": "",
+        "files": [
+          "**/*"
+        ]
```
MattIPv4, line 38:
```suggestion
        "basePath": "dist/zonos-elements",
        "files": [
          "*.@(css|js)"
        ]
```
Please avoid using a globstar as it can lead to including files that are not meant to be on the CDN. I suspect you want to include specifically the JS + CSS files under `dist/zonos-elements`. (And given that, I suspect you'll want the `filename` property in this JSON file, which defines the default file recommended, to be `zonos-elements.esm.js`).

MattIPv4 (changes requested):

packages/z/zonos-elements.json
Thienhuynh95: Thanks for the suggestion.

packages/z/zonos-elements.json
Thienhuynh95: I'll take these out, thanks.

packages/z/zonos-elements.json
Thienhuynh95: Thank you for your feedback. My default file for the CDN to load is `dist/scripts/loadZonos.js`, which I have specified in the `filename` at the root. This will then load all the necessary files from the `dist/zonos-elements` folder as required. I will make the adjustments and ask for your review.

Thienhuynh95:
@MattIPv4, I've implemented your suggestions. Please review and let me know if there's anything that requires further attention. Thank you.

MattIPv4 (approved):
LGTM

PR #1862, opened 2024-08-07
Update intl-tel-input w/ git auto-update

jackocnr:
I'm the creator of this package and have updated a few fields here that were out-of-date. Thanks.

### DIFF
--- a/packages/i/intl-tel-input.json
+++ b/packages/i/intl-tel-input.json
@@ -1,6 +1,6 @@
 {
   "name": "intl-tel-input",
-  "filename": "js/intlTelInput.min.js",
+  "filename": "build/js/intlTelInput.min.js",
   "homepage": "https://intl-tel-input.com",
   "description": "A JavaScript plugin for entering and validating international telephone numbers",
   "keywords": [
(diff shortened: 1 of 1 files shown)


packages/i/intl-tel-input.json
```diff
@@ -26,18 +26,15 @@
       {
         "basePath": "build",
         "files": [
-          "js/utils.js",
-          "js/intlTelInput*.js",
-          "img/flags*.png",
-          "css/intlTelInput.css"
+          "**/*"
```
MattIPv4, line 29:
```suggestion
          "css/*.css",
          "js/*.js",
          "js/i18n/{,*/}*.js",
          "img/*.@(webp|png)"
```
Please avoid using an unrestricted globstar as it can lead to including files on the CDN that are not supported.

MattIPv4 (changes requested):

jackocnr:
@MattIPv4 thanks for your input. I've made another change, as I realised I forgot the react build files. Does my change look ok to you? e.g. `"build/js/*.@(js|d.ts)"` - will this match files ending .js and .d.ts?

jackocnr:
@MattIPv4 I fixed the workflow issue.

MattIPv4 (approved):

PR #1865, opened 2024-08-13
Update css-scope-inline w/ git auto-update

jph00:
Apologies, but I used the wrong filename in css-scope-inline. This corrects it.

### DIFF
--- a/packages/c/css-scope-inline.json
+++ b/packages/c/css-scope-inline.json
@@ -1,6 +1,6 @@
 {
   "name": "css-scope-inline",
-  "filename": "script.min.js",
+  "filename": "script.js",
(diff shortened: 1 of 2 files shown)


MattIPv4:
👋 @jph00 I'll take a look at this in a bit (or a big bit), but also wanted to flag that you seem to have a version that is just published as `v`, which I suspect is going to trip up the automation: https://github.com/AnswerDotAI/css-scope-inline/releases/tag/v (it looks like [no versions have been added to the CDN](https://github.com/cdnjs/logs/tree/prod/packages/c/css-scope-inline) for it yet)

jph00:
Thanks for letting me know -- I didn't realise that would cause issues;
deleted it now.

MattIPv4 (approved):

PR #1894, opened 2024-09-25
Update tabler-icons w/ npm auto-update

totoprayogo1916:
tabler-icons is change path to @tabler/icons-webfont

### DIFF
--- a/packages/t/tabler-icons.json
+++ b/packages/t/tabler-icons.json
@@ -24,7 +24,8 @@
       {
         "basePath": "dist",
         "files": [
-          "**/!(*.html|*.scss)"
+          "fonts/*.@(ttf|woff|woff2)",
+          "*.css"
         ]
       }
     ]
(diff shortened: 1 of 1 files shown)


packages/t/tabler-icons.json
```diff
@@ -19,20 +19,17 @@
...
     "fileMap": [
       {
-        "basePath": "",
+        "basePath": "dist",
         "files": [
-          "tabler-*.svg",
-          "icons/*.svg",
-          "iconfont/tabler-*.@(css|scss)",
-          "iconfont/fonts/tabler-*"
+          "**/!(*.html|*.scss)"
```
MattIPv4, line 27:
```suggestion
          "fonts/*.@(ttf|woff|woff2)",
          "*.css"
```
Please avoid using globstars as they can result in files being added to the CDN that we don't want.

MattIPv4 (changes requested):

MattIPv4 (approved):
LGTM

Issue #1899, opened 2024-10-05
Add: sqids


4kimov:
### Library name

sqids

### Library license

https://github.com/sqids/sqids-javascript/blob/main/LICENSE

### Library source

https://www.npmjs.com/package/sqids

### Files to include

- cjs/sqids.js
- cjs/sqids.js.map
- cjs/sqids.d.ts
- esm/sqids.js
- esm/sqids.js.map
- esm/sqids.d.ts
- esm/package.json
- dist/sqids.js
- dist/sqids.js.map

PR #1903, opened 2024-10-08
Add sqids w/ npm auto-update

uiolee:
Resolves https://github.com/cdnjs/packages/issues/1899

### DIFF
--- a/packages/s/sqids.json
+++ b/packages/s/sqids.json
@@ -18,12 +18,12 @@
     "target": "sqids",
     "fileMap": [
       {
-        "basePath": "cjs",
-        "files": ["*.js", "*.map", "*.ts"]
+        "basePath": "",
+        "files": ["cjs/*.js", "cjs/*.map", "cjs/*.ts"]
       },
       {
-        "basePath": "esm",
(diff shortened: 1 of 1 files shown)


packages/s/sqids.json
```diff
@@ -0,0 +1,35 @@
...
+        "files": ["*.js", "*.map", "*.ts"]
+      },
+      {
+        "basePath": "esm",
+        "files": ["*.js", "*.map", "*.ts", "package.json"]
+      },
+      {
+        "basePath": "dist",
+        "files": ["*.js", "*.map"]
+      }
```
MattIPv4, line 31: Files in these directories have the same names, so they're going to overwrite each other when pulled out of the directories. If you're wanting to include the cjs/esm versions alongside the dist version, I'd recommend leaving `dist` with a `basePath` so that we pull that up, but then use an empty `basePath` for the cjs/esm versions so they stay in their subdirectory.

MattIPv4 (changes requested):

MattIPv4 (approved):
LGTM

PR #1929, opened 2024-12-05
Update jose w/ npm auto-update

panva:
As I'm gearing up for v6.x release this PR here updates the metadata to reflect the other package managers.

I'm also adding another glob from `dist/webapi` where v6.x will include its .js files. I am adding this instead just changing it because v4.x and v5.x may still be getting security fixes after v6.x release and those will remain in their existing (`dist/browser`) path.

### DIFF
--- a/packages/j/jose.json
+++ b/packages/j/jose.json
@@ -1,15 +1,21 @@
 {
   "name": "jose",
   "filename": "index.js",
-  "description": "'JSON Web Almost Everything' - JWA, JWS, JWE, JWT, JWK, JWKS for Node.js, Browser, Cloudflare Workers, Deno, Bun, and other Web-interoperable runtimes",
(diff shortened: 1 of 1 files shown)


packages/j/jose.json
```diff
@@ -57,6 +71,12 @@
         "files": [
           "**/*.js"
         ]
+      },
+      {
+        "basePath": "dist/webapi",
```
MattIPv4, line 76: Is this directory going to have files in it that have names that collide with files in dist/browser? As you're using `basePath` here, we're going to pull up all files from this subdirectory into the root

panva: the two basepaths are exclusive to their respective release lines

`<  v6` uses `dist/browser`
`>= v6` uses `dist/webapi`

MattIPv4: Ah, wonderful!

MattIPv4 (approved):

PR #1937, opened 2024-12-26
Add poplog w/ git auto-update

VArtzy:
PopLog is a simple, effective and lightweight JavaScript package that transforms console.log outputs into browser-friendly toast notifications. It makes debugging more interactive by displaying log messages as toasts on the screen, without the need to open developer tools.


MattIPv4 (changes requested):
Hi there!

This library doesn't have enough popularity (like more than 200 stars on GitHub or 800 downloads per month on npm). Please take a look at our [Policy, rules and guidelines](https://github.com/cdnjs/packages/blob/master/CONTRIBUTING.md#policy-rules-and-guidelines) for more information regarding this. 📖

Please open an issue or PR again when the library has more popularity, so that we can process it properly and add the library to cdnjs! Thanks for your great work and contribution, sorry we couldn't add your request at this time. 😄

VArtzy:
Thanks for your response, how do i find monthly download? I see my weekly download 1.078 on npm. Do I have wait for a month?

MattIPv4:
Yes, it would be good to see that level of downloads sustained for at least a month or so -- you'd also used GitHub as the auto-update mechanism here, so the lack of stars on the repository was also a factor.

Issue #1938, opened 2024-12-26
Add: clerk-js


Kshao123:
### Library name

@clerk/clerk-js

### Library license

https://github.com/clerk/javascript/blob/main/packages/clerk-js/LICENSE

### Library source

https://www.npmjs.com/package/@clerk/clerk-js

### Files to include

dist/*

PR #1939, opened 2024-12-26
Add clerk-js w/ npm auto-update

Kshao123:
#1938


Kshao123:
#1939

MattIPv4 (changes requested):
👋 Please take a look at the CI errors -- you need an `authors` array instead of an `author` object, and `repository.directory` is not a property we support.

Also, I think it might be more sensible to name this `clerk-js` rather than `clerk-clerk-js`?

MattIPv4 (approved):

PR #1950, opened 2025-01-30
Add ag-grid-enterprise w/ npm auto-update

JSkimming:


LBalsa (approved):

packages/a/ag-grid-enterprise.json
```diff
@@ -0,0 +1,49 @@
...
+    "grid",
+    "table",
+    "react",
+    "angular",
+    "angular-component",
+    "react-component",
+    "angularjs",
+    "reactjs"
+  ],
+  "license": "MIT",
```
MattIPv4, line 26: This does not seem correct?

MattIPv4 (changes requested):

packages/a/ag-grid-enterprise.json
JSkimming: I've updated it say `Commercial see LICENSE.txt`
I based it on the example similar to [datatables.net-editor-bs4.json](https://github.com/cdnjs/packages/blob/master/packages/d/datatables.net-editor-bs4.json#L9).

MattIPv4 (approved):

PR #1972, opened 2025-03-18
Add typescript-vfs w/ npm auto-update

SukkaW:
freeCodeCamp is trying to migrate from a npm-based CDN to cdnjs (https://github.com/freeCodeCamp/freeCodeCamp/pull/59291). `@typescript/vfs`, the Map-based TypeScript Virtual File System commonly used for running TypeScript in-browser, is the last piece of the puzzle missing in cdnjs.

The npm weekly download stat of the package is now 600,000+.

cc @huyenltnguyen

### DIFF
--- a/packages/t/typescript-vfs.json
+++ b/packages/t/typescript-vfs.json
@@ -19,7 +19,7 @@
       {
         "basePath": "dist",
         "files": [
-          "**/*"
+          "*.js"
         ]
       }
     ]
--- a/packages/t/typescript-vfs.json
+++ b/packages/t/typescript-vfs.json
@@ -1,4 +1,5 @@
 {
+  "name": "typescript-vfs",
   "description": "A Map based TypeScript Virtual File System.",
   "filename": "vfs.globals.js",
   "homepage": "http://www.typescriptlang.org",
@@ -7,7 +8,6 @@
     "vfs"
   ],
   "license": "MIT",
-  "name": "typescript",
   "repository": {
     "type": "git",
     "url": "https://github.com/microsoft/TypeScript-Website"
--- a/packages/t/typescript-vfs.json
+++ b/packages/t/typescript-vfs.json
@@ -1,6 +1,6 @@
 {
   "description": "A Map based TypeScript Virtual File System.",
-  "filename": "dist/vfs.cjs.production.min.js",
+  "filename": "dist/vfs.globals.js",
   "homepage": "http://www.typescriptlang.org",
   "keywords": [
     "typescript",
--- a/packages/t/typescript-vfs.json
+++ b/packages/t/typescript-vfs.json
@@ -22,10 +22,6 @@
           "**/*"
         ]
       }
-    ],
-    "ignoreVersions": [
-      "*dev*",
-      "*insiders*"
     ]
   },
   "authors": [
--- a/packages/t/typescript-vfs.json
+++ b/packages/t/typescript-vfs.json
@@ -1,6 +1,6 @@
 {
   "description": "A Map based TypeScript Virtual File System.",
-  "filename": "dist/vfs.globals.js",
+  "filename": "vfs.globals.js",
   "homepage": "http://www.typescriptlang.org",
   "keywords": [
     "typescript",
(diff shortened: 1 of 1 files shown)


MattIPv4 (commented):
👋 This library makes extensive usage of `require` which does not work in the browser. As you sure you want to add this to the CDN?

packages/t/typescript-vfs.json
```diff
@@ -0,0 +1,32 @@
+{
+  "description": "A Map based TypeScript Virtual File System.",
+  "filename": "dist/vfs.globals.js",
```
MattIPv4, line 3:
```suggestion
  "filename": "vfs.globals.js",
```
As you've set basePath to dist, you don't need to include it here

packages/t/typescript-vfs.json
```diff
@@ -0,0 +1,32 @@
...
+    "url": "https://github.com/microsoft/TypeScript-Website"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "@typescript/vfs",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 22:
```suggestion
          "*.js"
```
We try to avoid adding globstars, we'd prefer a glob that is more specific to the actual files that're needed.

SukkaW:
> 👋 This library makes extensive usage of `require` which does not work in the browser. As you sure you want to add this to the CDN?

IIRC, the `vfs.globals.js` uses IIFE and no CJS variables (`exports`, `module`, `require`) so it is safe to be used in the browsers. Heck, they even reference the `localStorage`.

There is also `vfs.esm.js` which can be used with `<script module />`.

MattIPv4:
```
cat node_modules/@typescript/vfs/dist/vfs.globals.js| grep require
  var path = requirePath();
  var fs = requireFS();
    var lib = tsLibDirectory || path.dirname(require.resolve("typescript"));
  var libFiles = fs.readdirSync(tsLibDirectory || path.dirname(require.resolve("typescript")));
  var path = requirePath();
  var fs = requireFS();
  var path = requirePath();
  var tsLib = tsLibDirectory != null ? tsLibDirectory : path.dirname(require.resolve("typescript"));
var requirePath = function requirePath() {
  return require(String.fromCharCode(112, 97, 116, 104));
var requireFS = function requireFS() {
  return require(String.fromCharCode(102, 115))
```

These seem to be direct calls to `require` without it being overridden by anything like how you'd see in Webpack?

The ESM version seems to make the same `require` calls -- have you confirmed this works in a browser without additional bundling?

```
cat node_modules/@typescript/vfs/dist/vfs.esm.js| grep require    
  var path = requirePath();
  var fs = requireFS();
    var lib = tsLibDirectory || path.dirname(require.resolve("typescript"));
  var libFiles = fs.readdirSync(tsLibDirectory || path.dirname(require.resolve("typescript")));
  var path = requirePath();
  var fs = requireFS();
  var path = requirePath();
  var tsLib = tsLibDirectory != null ? tsLibDirectory : path.dirname(require.resolve("typescript"));
var requirePath = function requirePath() {
  return require(String.fromCharCode(112, 97, 116, 104));
var requireFS = function requireFS() {
  return require(String.fromCharCode(102, 115));
```

SukkaW:
> These seem to be direct calls to `require` without it being overridden by anything like how you'd see in Webpack?

There aren't any direct `require()` calls. Those are already transformed by `@rollup/commonjs-plugin` and ready to be run on browsers.

SukkaW:
@MattIPv4 Check out https://sukka-reproductions.github.io/cdnjs-1972/

This demonstrates that the TypeScript VFS is working in the browsers.

See also `https://www.typescriptlang.org/js/9d44735/sandbox/vendor/typescript-vfs.js` used in https://www.typescriptlang.org/play/, it uses the AMD version of the vfs and it also contains `require` call, yet the TypeScript playground just works.

MattIPv4:
> A JavaScript file containing `require()` doesn't mean it won't run on browsers, instead those `require()` might be dead code and will never be reached.

I am aware of that :) The methods that they are called in are exported, so I assume they were not dead code. I trust your assessment that they are in fact not needed 👍

MattIPv4 (approved):

PR #2011, opened 2025-05-29
Add temporal-polyfill w/ npm auto-update

Keritial:
This PR adds `temporal-polyfill` to cdnjs.

Temporal is shipped in Firefox 139, but not in other major browsers including Chrome and Safari.

It has 266934 weekly downloads which satisfies the requirement.

### DIFF
--- a/packages/t/temporal-polyfill.json
+++ b/packages/t/temporal-polyfill.json
@@ -1,0 +1,31 @@
+{
+  "name": "temporal-polyfill",
+  "description": "A lightweight polyfill for Temporal, successor to the JavaScript Date object",
+  "filename": "global.min.js",
+  "keywords": [
+    "Temporal",
+    "polyfill",
+    "date",
+    "time"
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/fullcalendar/temporal-polyfill"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "temporal-polyfill",
+    "fileMap": [
+      {
+        "basePath": "",
+        "files": [
+          "**/*.js"
+        ]
+      }
+    ]
+  },
+  "optimization": {
+    "js": false
+  }
+}
--- a/packages/t/temporal-polyfill.json
+++ b/packages/t/temporal-polyfill.json
@@ -20,7 +20,11 @@
       {
         "basePath": "",
         "files": [
-          "global*.@(js|cjs|d.ts|d.cts)"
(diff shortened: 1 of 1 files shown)


MattIPv4 (changes requested):

Keritial:
Thanks for your review.

I have reconsidered with the fact that cdnjs is primarily a CDN for browsers, we should focus on distributing only files needed for browsers.

This means excluding Node-specific files like `.cjs`, `.cts` and TypeScript definitions (`.d.ts`, `.d.cts`) that aren't required in browsers.

For browser environment, there are two ways to use this library:

1. Global script tag: By including `global.js` via `<script>` tags
```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/temporal-polyfill/0.3.0/global.min.js"></script>
```
This will add Temporal to the global object (`window`). Then we can use Temporal elsewhere in the code.

2. ES Module: Allowing users to import via:
```html
<script type="module">
	import { Temporal } from "https://cdnjs.cloudflare.com/ajax/libs/temporal-polyfill/0.3.0/index.js";
	console.log(Temporal.Now.instant();
</script>
```
`index.js` imports files under the `chunk` folder, so they are included too.

Keritial:
By the way, `chunk/classApi.js` and `chunk/internal.js` is not minified. Should we set `optimization.js` to `true`? Will cdnjs replace the import in `index.js` to the minified version?

packages/t/temporal-polyfill.json
```diff
@@ -0,0 +1,35 @@
...
+    "target": "temporal-polyfill",
+    "fileMap": [
+      {
+        "basePath": "",
+        "files": [
+          "global.min.js",
+          "global.esm.js",
+          "index.js",
+          "chunk/classApi.js",
+          "chunk/internal.js"
```
MattIPv4, line 27:
```suggestion
          "chunk/*.js"
```
I think we're safe to use `*` here so handle any other future chunks being imported

packages/t/temporal-polyfill.json
```diff
@@ -0,0 +1,34 @@
...
+    "source": "npm",
+    "target": "temporal-polyfill",
+    "fileMap": [
+      {
+        "basePath": "",
+        "files": [
+          "global.min.js",
+          "global.esm.js",
+          "index.js",
+          "chunk/*.js"
```
MattIPv4, line 26:
```suggestion
          "chunks/*.js"
```

MattIPv4 (approved):

Issue #1803, opened 2024-04-24
Add: elkjs


dbusteed:
### Library name

elkjs

### Library license

https://github.com/kieler/elkjs/blob/master/LICENSE.md

### Library source

https://www.npmjs.com/package/elkjs

### Files to include

lib/

PR #2021, opened 2025-06-24
Add mermaid-layout-elk w/ npm auto-update

bollwyvl:
As always, thanks for `cdnjs`!

This adds `@mermaid-js/layout-elk`, which was unvendored from [mermaid](https://github.com/cdnjs/packages/blob/master/packages/m/mermaid.json) somewhere along the line. 

It is ESM-only, and looks like it vendors [elkjs](https://github.com/cdnjs/packages/issues/1803), if that causes any concerns.

### DIFF
--- a/packages/m/mermaid-layout-elk.json
+++ b/packages/m/mermaid-layout-elk.json
@@ -18,7 +18,8 @@
       {
         "basePath": "dist",
         "files": [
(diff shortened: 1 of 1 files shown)


MattIPv4 (approved):

bollwyvl:
Thanks!

PR #2033, opened 2025-08-08
Add xom-sharepoint w/ npm auto-update

juliolmuller:


packages/x/xom-sharepoint.json
```diff
@@ -0,0 +1,38 @@
...
+    "url": "https://github.com/juliolmuller/xom-sharepoint"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "xom-sharepoint",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 33:
```suggestion
          "*.js"
```
We don't accept open globstars in configs as it can lead to files be included on the CDN that we don't expect. It looks like there are just a couple of JS files in the dist directory that are needed?

MattIPv4 (approved):

PR #2034, opened 2025-08-08
Add file2arraybuffer w/ npm auto-update

juliolmuller:


packages/f/file2arraybuffer.json
```diff
@@ -0,0 +1,37 @@
...
+    "url": "https://github.com/juliolmuller/file2arraybuffer-js"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "file2arraybuffer",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 32:
```suggestion
          "*.js"
```
We don't accept open globstars in configs as it can lead to files being included on the CDN that we don't expect. It looks like there are just a couple of JS files in the dist directory that are needed?

MattIPv4 (changes requested):

PR #2041, opened 2025-08-08
Add cpf-utils w/ npm auto-update

juliolmuller:


packages/c/cpf-utils.json
```diff
@@ -0,0 +1,43 @@
...
+    "url": "https://github.com/juliolmuller/cpf-utils-js"
+  },
+  "autoupdate": {
+    "source": "npm",
+    "target": "cpf-utils",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 38:
```suggestion
          "*.js"
```
We don't accept open globstars in configs as it can lead to files being included on the CDN that we don't expect. It looks like there are just a couple of JS files in the dist directory that are needed?

MattIPv4 (changes requested):

MattIPv4 (approved):

PR #2051, opened 2025-08-30
Add tailwindplus-elements w/ npm auto-update

JSkimming:


packages/t/tailwindplus-elements.json
```diff
@@ -0,0 +1,34 @@
...
+    }
+  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "@tailwindplus/elements",
+    "fileMap": [
+      {
+        "basePath": "dist",
+        "files": [
+          "**/*"
```
MattIPv4, line 29:
```suggestion
          "*.@(js|cjs|d.ts|d.cts)"
```
We don't accept adding open globstars as they can lead to including files not wanted on the CDN.

MattIPv4 (changes requested):

MattIPv4 (approved):

PR #2052, opened 2025-09-08
Update tailwindplus-elements w/ npm auto-update

JSkimming:
- [x] Remove `*.cjs` and `*.d.cts` as they are not whitelisted by CDN and therefore not served.
- [x] ~Exclude _insiders_ builds.~ - This caused the error `autoupdate: Additional property ignoreVersions is not allowed`
- [x] Prevent minification, as the `index.js` file is already minified.

### DIFF
--- a/packages/t/tailwindplus-elements.json
+++ b/packages/t/tailwindplus-elements.json
@@ -29,6 +29,9 @@
           "*.@(js|d.ts)"
         ]
       }
+    ],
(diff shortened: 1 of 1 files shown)


JSkimming:
I've found several other NPM packages using `ignoreVersions`. I've removed it as the other changes are worth having. Is there a way to exclude some NPM package versions?

PR #2099, opened 2025-12-17
Add aws-sdk3-client-scheduler w/ npm auto-update

1mnj1:
Add AWS SDK v3 for Client Scheduler package

Source: https://github.com/aws/aws-sdk-js-v3
NPM: https://www.npmjs.com/package/@aws-sdk/client-scheduler

### DIFF
--- a/packages/a/aws-sdk3-client-scheduler.json
+++ b/packages/a/aws-sdk3-client-scheduler.json
@@ -1,0 +1,35 @@
+{
+  "name": "aws-sdk3-client-scheduler",
+  "description": "AWS SDK3 for JavaScript - Client Scheduler",
+  "keywords": [
+    "api",
+    "amazon",
+    "aws",
+    "eventbridge",
+    "scheduler"
+  ],
+  "autoupdate": {
+    "source": "npm",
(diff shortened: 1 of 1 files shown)


packages/a/aws-sdk3-client-scheduler.json
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  "autoupdate": {
+    "source": "npm",
+    "target": "aws-sdk/client-scheduler",
+    "fileMap": [
+      {
+        "basePath": "dist-es", 
+        "files": [
+          "*.js",
+          "**/*.js"
```
MattIPv4: I'd like to avoid introducing a config that includes a globstar, as it can result in files being included that we don't need on the CDN. Can we make this just `*/*.js`?

packages/a/aws-sdk3-client-scheduler.json
```diff
@@ -0,0 +1,35 @@
...
+        ]
+      }
+    ]
+  }
```
MattIPv4:
```suggestion
  },
```

MattIPv4 (changes requested):

PR #2121, opened 2026-02-11
Add promptfont w/ npm auto-update

Shinmera:
The font features over 500 glyphs, many of which are specifically designed for use in games -- to present easy to read input prompts, and for use in user interface design.

PromptFont used to live at https://github.com/shinmera/promptfont, where it had over 200 stars, though it has moved to Codeberg since: https://codeberg.org/shinmera/promptfont

You can see a listing of all the glyphs along with an interactive site here: https://shinmera.com/docs/promptfont/

The cdnjs package only includes files that would be relevant for CDN use, namely the font files, the css file, and the glyphs metadata file. I hope that fits the intended use of cdnjs.

Thanks a lot for all your work on maintaining this stuff, I've been happily making use of it for many years! ❤️

### DIFF
--- a/packages/p/promptfont.json
+++ b/packages/p/promptfont.json
@@ -1,6 +1,6 @@
 {
   "name": "promptfont",
-  "filename": "css/all.min.css",
+  "filename": "promptfont.css",
   "description": "A font designed for button prompts and user interfaces in games.",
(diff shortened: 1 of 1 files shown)


Shinmera:
Thanks a lot!

PR #2144, opened 2026-04-01
Create regina.js

muslax:
Regina


MattIPv4 (changes requested):
You cannot add arbitrary JavaScript files to this repository. If you wish to add a library to cdnjs, you need to add a JSON config file for it with a valid auto-update map to a GitHub repository with tags or an NPM package.

PR #2150, opened 2026-04-15
Add ag-charts-community w/ npm auto update

seanlandsman:


packages/a/ag-charts-community.json
```diff
@@ -0,0 +1,40 @@
...
+  "license": "MIT",
+  "homepage": "http://www.ag-charts.com/charts",
+  "autoupdate": {
+    "source": "npm",
+    "target": "ag-charts-community",
+    "fileMap": [
+      {
+        "basePath": "dist/umd",
+        "files": [
+          "**/*"
```
MattIPv4:
```suggestion
          "*.js"
```
We like to avoid using open glob stars in the configs, as they can result in files being included that aren't meant for the CDN.

MattIPv4 (approved):

PR #2176, opened 2026-07-03
Update js-yaml w/ git auto-update

LenaWil:
The directory it puts the main library in seems to have changed. 

I couldn’t figure out how to do offline tests, so I hope this works first try. 
I am personally on version 4, so I hope this one takes the older versions too.


MattIPv4 (changes requested):
https://github.com/nodeca/js-yaml/blob/master/bin/js-yaml.mjs is not a valid script to have on the CDN, it relies on `node:` imports, and does not contain the library.

The git repository no longer contains the `dist` directory, I suspect you will want to change the `autoupdate` config to pull from the package's npm distribution instead.

MattIPv4:
No progress, closing.

PR #2181, opened 2026-08-03
Add maidr w/ npm auto-update

jooyoungseo:
Adds [maidr](https://maidr.ai) — Multimodal Access and Interactive Data Representation, an accessibility library that gives blind and low-vision users non-visual access to statistical charts through audio sonification, text descriptions, braille output, and keyboard navigation. Developed by the (x)Ability Design Lab at the University of Illinois Urbana-Champaign.

### Popularity

- **npm:** ~2,200 downloads/month ([maidr](https://www.npmjs.com/package/maidr)), above the 800/month baseline for npm-sourced libraries.
- Actively maintained, released via semantic-release, currently at v3.75.1.

### Licence

`GPL-3.0-or-later`, declared in both `package.json` and a root `LICENSE` file (verbatim GPL-3.0 text).

### Auto-update

npm-sourced, with three files listed explicitly rather than a glob, per [#186](https://github.com/cdnjs/packages/issues/186):

| file | size | why |
|---|---|---|
| `maidr.js` | ~1.9 MB | the library (UMD) |
| `maidr.css` | 406 B | near-empty on purpose; the interface is styled at runtime, and the file is still published so existing `<link rel="stylesheet">` tags don't 404 |
| `maidr-math.css` | ~360 KB | KaTeX stylesheet for LaTeX in AI chat responses, linked at runtime **relative to the URL `maidr.js` was loaded from** — so on a page served from cdnjs it resolves to a cdnjs URL, and omitting it would 404 |

A glob was deliberately avoided: `dist/` also contains ten chart-library adapter bundles at roughly 1.9 MB each plus sourcemaps, and `maidr.js.map` alone is 8.7 MB. None of that is needed to use the library from a CDN.

All three files are present in the published tarball — verified against `https://cdn.jsdelivr.net/npm/maidr@latest/dist/`.

### Checks run before submitting

- Validated against `schema_human.json` from `cdnjs/tools`: all required fields present (`name`, `description`, `keywords`, `repository`, `autoupdate`), no properties outside the schema, and `autoupdate` satisfies `fileMap`/`source`/`target` with `basePath` + non-empty `files`.
- No existing `maidr` library on cdnjs (`api.cdnjs.com/libraries/maidr` → 404).
- File committed with LF endings, matching the other files in `packages/m/`.

Happy to adjust the file map or any metadata if you'd prefer something different.


jooyoungseo:
Thanks — I checked this rather than assuming, and I believe the `filename` is correct as written, but I'm happy to change it if you'd prefer.

cdnjs generates the minified variants itself. From [CONTRIBUTING.md](https://github.com/cdnjs/packages/blob/master/CONTRIBUTING.md): *"By default, cdnjs automatically generates minified versions of JavaScript and CSS files … These optimizations can be disabled per-library using the `optimization` property."* This manifest doesn't set `optimization`, so the defaults apply.

Verified against a live library rather than relying on the docs alone — `marked@16.3.0`:

- its npm package contains `lib/marked.esm.js` and `lib/marked.umd.js`, and **no `.min.js` under `lib/`**
- its file map for that directory is `lib/*.@(cjs|js)`, which matches only those unminified files
- yet cdnjs serves `lib/marked.esm.min.js` and `lib/marked.umd.min.js` — `https://cdnjs.cloudflare.com/ajax/libs/marked/16.3.0/lib/marked.umd.min.js` returns `200`

So `maidr.js` being mirrored is what produces `maidr.min.js`. This is also the common shape here: 3,266 manifests in `packages/` declare a `.min.*` `filename` that their own file map doesn't list.

(I did check the one counter-example I found — `ably` 404s on its declared `ably.min.js` — but that's unrelated: its `assets` list is empty for the current version, so its `browser/static` base path isn't matching anything in the npm package at all. Nothing is being mirrored there, minified or otherwise.)

Two things worth flagging in case they affect your preference:

1. `dist/maidr.js` is **already minified** by the project's own build (941 lines, ~2,010 chars/line), so `maidr.min.js` will be a re-minification of minified code — functionally fine, just near-identical to the source file.
2. `dist/maidr.js` ends with `//# sourceMappingURL=maidr.js.map`, and the map is deliberately **not** mirrored — it's 8.7 MB on its own. So devtools will 404 on the map when opened. I can strip nothing from my side, but if you'd rather the listing not carry a dangling map reference, say so and I'll look at fixing it upstream in the package instead.

If you'd prefer `"filename": "maidr.js"` — matching what the project's own docs, [py-maidr](https://py.maidr.ai/) and [r-maidr](https://r.maidr.ai/) all link — I'm glad to switch it. Just let me know which you'd rather have.

MattIPv4 (approved):

PR #2189, opened 2026-08-12
canvasXpress: fix stalled autoupdate (git:// protocol removed by GitHub)

neuhausi:
The autoupdate target used the git:// protocol, which GitHub disabled in 2022, so autoupdate stalled at 24.8.1. Switching to https:// restores autoupdate. The fileMap basePath (inst/htmlwidgets/lib/canvasXpress) still resolves at current tags — verified canvasXpress.css and canvasXpress.min.js are present at tag 65.6.


MattIPv4:
I don't believe this change is needed, we don't use the URL directly, we parse it. Updates are being processed for this library just fine: https://github.com/cdnjs/logs/tree/prod/packages/c/canvasXpress

PR #2197, opened 2026-09-01
Update docsify w/ npm auto-update

sy-records:


sy-records:
@MattIPv4 Hi, Could you review it?

packages/d/docsify.json
```diff
@@ -20,9 +20,12 @@
...
     "fileMap": [
       {
-        "basePath": "lib",
+        "basePath": "",
         "files": [
-          "**/*"
+          "dist/**/*",
+          "lib/**/*",
+          "src/**/*",
+          "themes/*"
```
MattIPv4: We try to avoid landing globstars in new configs, as it can pull files onto the CDN that aren't needed/wanted. Please can you make these more specific to the file extensions + directories that need to be included.

Why is the whole of `src` and `themes` needed when `dist` + `lib` seem to have the same files?

MattIPv4 (changes requested):

packages/d/docsify.json
sy-records: Okay, I'll edit it later.

There are duplicate directories because, in the past, too many users did not specify a version. To avoid affecting users, we have made v5 backward compatible with v4.

packages/d/docsify.json
```diff
@@ -20,9 +20,12 @@
...
     "fileMap": [
       {
-        "basePath": "lib",
+        "basePath": "",
         "files": [
-          "**/*"
+          "dist/**/*.{js,css,js.map,css.map}",
+          "lib/**/*.{js,css}",
+          "src/**/*.{js,ts,css,js.map,css.map,d.ts.map}",
+          "themes/**/*.css"
```
sy-records: Or do I need to provide the complete list?

```suggestion
          "dist/*.{js,js.map}",
          "dist/plugins/*.{js,js.map}",
          "dist/plugins/assets/*.css",
          "dist/themes/*.{css,css.map}",
          "dist/themes/addons/*.{css,css.map}",
          "lib/*.js",
          "lib/plugins/*.js",
          "lib/themes/*.css",
          "src/core/*.{js,ts,d.ts.map}",
          "src/core/event/*.{js,d.ts,d.ts.map}",
          "src/core/fetch/*.{js,d.ts,d.ts.map}",
          "src/core/init/*.{js,d.ts,d.ts.map}",
          "src/core/render/*.{js,d.ts,d.ts.map}",
          "src/core/render/compiler/*.{js,d.ts,d.ts.map}",
          "src/core/router/*.{js,d.ts,d.ts.map}",
          "src/core/router/history/*.{js,d.ts,d.ts.map}",
          "src/core/util/*.{js,d.ts,d.ts.map}",
          "src/core/virtual-routes/*.{js,d.ts,d.ts.map}",
          "src/plugins/*.{js,d.ts,d.ts.map}",
          "src/plugins/front-matter/*.{js,d.ts,d.ts.map}",
          "src/plugins/search/*.{js,css,d.ts,d.ts.map}",
          "src/themes/*.css",
          "src/themes/addons/*.css",
          "src/themes/shared/*.css",
          "themes/*.css"
```

packages/d/docsify.json
MattIPv4: > There are duplicate directories because, in the past, too many users did not specify a version. To avoid affecting users, we have made v5 backward compatible with v4.

cdnjs _requires_ that users specify a version, so that does not make any sense in the context of the configuration for cdnjs -- please avoid duplicates in the config, only include the resources actually needed.

packages/d/docsify.json
MattIPv4: As above, I do not follow why any of `src` is needed...?

packages/d/docsify.json
sy-records: Now only the “dist” directory is retained.

trusktr: @MattIPv4 we publish src for reasons on npmjs.com, which automatically propagates to other CDNs (f.e. JSDelivr). See the file structure here:

https://www.npmjs.com/package/docsify?activeTab=code

It would be ideal that our published source on any CDN always matches with npm (ie our package.json `files` config) everywhere, and to avoid having special cases per CDN. The `"src/**/*"`, etc, above was recreating our `files` config:

```js 
  "files": [
    "dist",
    "src",
    "lib",
    "themes"
  ],
```

I'm not experienced with CDNjs. Is there a better way to apply the same `files` from our package.json?

packages/d/docsify.json
```diff
@@ -32,6 +32,11 @@
       "name": "qingwei-li",
       "email": "cinwell.li@gmail.com",
       "url": "https://github.com/QingWei-Li"
+    },
+    {
+      "name": "Luffy",
+      "email": "lufei@docsifyjs.org",
+      "url": "https://github.com/sy-records"
```
trusktr, line 42: Can we put a generic Docsify as author here rather than specific people, pointing to GitHub.com/docsify? That might be better so anyone with questions can easily find the repo.

QingWei-Li left the project so having him there is probably not helpful.

packages/d/docsify.json
trusktr: Explained `src` in the other comment 

- https://github.com/cdnjs/packages/pull/2197#discussion_r4023481624

packages/d/docsify.json
MattIPv4: Sure, a generic author for the project sounds fine 👍

packages/d/docsify.json
MattIPv4: cdnjs operates differently from other CDNs -- we very intentionally do not just proxy all package files through by default, as the vast majority of them tend not to be relevant to consuming a library on a website frontend.

cdnjs does not do any processing to the files beyond minifying them, so I find it unlikely that `src` is going to be useful to folks?

You need to provide an explicit mapping of files in the cdnjs JSON file being introduced in this PR, it will not automatically pull from your `files` nor should it.

Please only include the assets relevant to the CDN itself for folks consuming this library on websites, which presumably would be the bundled files in your `dist` directory?

trusktr: Src has plain JavaScript modules and existed before a dist module. For some sites if cndjs doesn't have src and they swap jsdelivr with cdnjs in their importmap, for example, those sites won't work.

Some tools also support installing from JS modules via importmaps (files can come from anywhere). Those tools allow for installing with support for Go To Definition in IDEs with type checking enabled, straight from CDN. Deno is one example.

We would like to support full breadth of options.

Also our files are meant to be backwards compatible. Dist/ is v5. Lib/ and themes/ are v4. 

We'd like to keep consistency in how we publish our lib.

MattIPv4: 👍 Okay, if you insist. We avoid globstars in configs, though, so you'll need to list out specific directories and the file extensions that need to be kept.

MattIPv4 (approved):

PR #2198, opened 2026-09-02
canvasXpress: fix autoupdate target (git:// → https://)

neuhausi:
The autoupdate.target uses the git:// protocol, which GitHub permanently disabled in 2022. As a result the auto-updater can no longer clone the repo and the library has been frozen at 24.8.1 while upstream is at 67.5. This one-line change switches the target to https://; tags and the fileMap path are already correct, so the updater should backfill the missing versions on the next run.

### DIFF
--- a/packages/c/canvasXpress.json
+++ b/packages/c/canvasXpress.json
@@ -20,7 +20,7 @@
   },
   "autoupdate": {
     "source": "git",
-    "target": "git://github.com/neuhausi/canvasXpress.git",
+    "target": "https://github.com/neuhausi/canvasXpress.git",
     "fileMap": [
       {
         "basePath": "inst/htmlwidgets/lib/canvasXpress",
@@ -38,3 +38,4 @@
     }
   ]
 }
+


MattIPv4:
As I explained in https://github.com/cdnjs/packages/pull/2189#issuecomment-5270170891, we do not use the target directly, we parse it. Your library is being updated on cdnjs just fine... https://cdnjs.com/libraries/canvasXpress/67.0 is the most recent version ([from 9 hours ago](https://github.com/cdnjs/logs/commit/c1c5a46668aed07e03faa09d9837035ba7baf415)) we have, and I imagine the bot will pick up newer ones in due course.

neuhausi:
Please accept my PR. 

Here is what I get when I go to https://cdnjs.com/libraries?search=canvasxpress -> It show 24.8.1

Untitled

Also in my GitHub badge:

Untitled

MattIPv4:
Your PR will not change anything.

24.8.1 shows as the latest version because when sorted by semver, it is the latest version.

Your recent versions are not 3-part semver, and so sort as older versions.

neuhausi:
Ooops. Thank you

neuhausi:
Thanks for the semver explanation, that makes sense. We've since retagged the recent releases in proper 3-part semver — 67.6.0, 67.7.0, and 67.8.0 are now on GitHub (and have been for a few days).

However, cdnjs still hasn't imported any of them — the latest version it shows is 67.5.zenodo, and the badge is still resolving to 24.8.1. It also skipped the 67.5 and 67.6 tags from the same batch. So it looks like autoupdate hasn't run (or has stalled) for canvasXpress since 67.5.zenodo.

Could you trigger a manual re-import for the package? Once 67.8.0 lands it should sort correctly as the latest. Thanks!

neuhausi:
Thank you so much. It works now

